PR #1, opened 2020-01-29
Minor CSS fixes with the viz container and export button alignment.

jrdzha:


dorisjlee:
The bug fixes look good. Is there a way to do the PR without the commit 1a97437 (merge branch)?

PR #7, opened 2020-05-31
Extend Parser to support exclusion as Spec field and associated tests

caitlynachen:


dorisjlee:
This looks great! Thanks @caitlynachen!

PR #8, opened 2020-06-01
add filter attribute to View repr

westernguy2:


dorisjlee:
The structure of the _repr_ looks good, although it is not necessary to create a new filter_attribute and filter_value in `Filter.py`. Please remove the creation of `filter_attribute` and `filter_value` in `Filter.py` and leverage the `attribute` and `value` in the `Spec` object directly to display the filter value.

dorisjlee:
Squashed and merged into master. Thanks @westernguy2 !

PR #9, opened 2020-06-01
Pandas Executor and Interestingness Bugfix

19thyneb:
When dealing with filtered aggregation with missing groupby-attribute values, issue in cases where more than 1 aggregated value which caused a column mismatch. Executor now handles this case appropriately.

Issue in interestingness function where v_filter and v are of different lengths, now samples v_filter to make it the same size and the unfiltered value set

### DIFF
--- a/lux/action/Correlation.py
+++ b/lux/action/Correlation.py
@@ -38,7 +38,21 @@ def correlation(ldf:LuxDataFrame,ignoreTranspose:bool=False):
 	# if (ignoreIdentity): vc = filter(lambda x: x.specLst[0].attribute!=x.specLst[1].attribute,ldf.viewCollection)
 	vc = Compiler.compile(ldf, vc, enumerateCollection=False)
 
+	#for benchmarking executor
+	if ldf.toggleBenchmarking == True:
+		ticExec = time.perf_counter()
 	ldf.executor.execute(vc,ldf)
+	if ldf.toggleBenchmarking == True:
+		import pandas as pd
+		tocExec = time.perf_counter()
+		benchmarkData = {'action': ['correlation'], 'executor_type': [ldf.executorType], 'action_phase': ['viewcollection_execution'], 'time': [tocExec-ticExec]}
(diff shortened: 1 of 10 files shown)


dorisjlee:
Thanks @thyneb19, I agree this is an important bug to fix. I tried to fix this in commit [18f4a6a](https://github.com/lux-org/lux/commit/18f4a6a11831fe956595ecbf465afbf8f54c5ae7) and removed the sample code then. I didn't quite understand the logic to sample v_filter to make v and v_filter the same size.  My impression was that padding of the zeroes in the PandasExecutor would be sufficient to fix this issue. Maybe I'm missing something and you could help explain. 
It might also be useful for @caitlynachen to look over her code for zero padding to see if the changes make sense and why the unequal length issue is still happening. 
@thyneb19 : Could you post the example that you described above (filter aggregation with missing group by), so we could use it for testing?

19thyneb:
So the bug within the interestingness function actually only happens when using the SQL Executor not with the Pandas Executor, so this might mean that we just need to add the zero padding code to the SQL Executor.

For the filter aggregation bug, I was setting the context to be [depaturedelay, month]. When a filter was applied, a row in the view.data should have 3 values associated with it (one for the filter variable value, and a 0 for both depaturedelay and month) but only 2 values were being added (one for the filter variable value, and a single 0).

dorisjlee:
I think this means that we need to add zero padding for the SQLExecutor. 
Not sure if I completely understand why there is three values after the filter on [depaturedelay, month], shouldn't there only be 2 values one for depaturedelay and one for month. It would be useful to isolate which filter example is causing this so we can understand how to pad the zero appropriately in the SQLExecutor.

dorisjlee:
I'm going to close this PR for now, since the issue might have been resolved with commit [18f4a6a](https://github.com/lux-org/lux/commit/18f4a6a11831fe956595ecbf465afbf8f54c5ae7). If there is something we want to merge in here for the SQLExecutor, let's start a separate PR for this and base it off the updated code to avoid merge conflicts.

PR #10, opened 2020-06-03
Adding logic for removing invalid views in compiler, and updated compiler tests. 

jaywoo123:
Adding code in Complier.py that adds a step for removing invalid views during view collection generation(ex. year by year). Adding old test cases back to test_compile.py and adding showMore calls to the test cases.

### DIFF
--- a/lux/compiler/Compiler.py
+++ b/lux/compiler/Compiler.py
@@ -50,6 +50,9 @@ class Compiler:
 		# Output : DataObj/DataObjectCollection
 		# compiledCollection = []
 		viewCollection = Compiler.expandUnderspecified(ldf, viewCollection)  # autofill data type/model information
+
+		viewCollection = Compiler.removeAllInvalid(viewCollection)
+
 		for view in viewCollection:
 			Compiler.determineEncoding(ldf, view)  # autofill viz related information
 		return viewCollection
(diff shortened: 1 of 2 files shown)


lux/compiler/Compiler.py
```diff
@@ -111,6 +114,8 @@ def expandUnderspecified(ldf, viewCollection):
 		ldf : lux.luxDataFrame.LuxDataFrame
 			LuxDataFrame with underspecified context
 
+		viewCollection : list[lux.view.View]
+			empty list that will be populated with specified lux.View objects.
```
dorisjlee, line 118: This is not always empty right?

lux/compiler/Compiler.py
```diff
@@ -136,6 +141,35 @@ def expandUnderspecified(ldf, viewCollection):
 						chartTitle = spec.value
 					view.title = f"{spec.attribute} {spec.filterOp} {chartTitle}"
 		return views
+
+	@staticmethod
+	def removeAllInvalid(viewCollection):
+		"""
+		Given an expanded view collection, remove all views that are invalid.
```
dorisjlee, line 148: Add an explanation regarding what types of views are considered invalid and what we are removing here.

jaywoo123: Added to a new commit.

lux/compiler/Compiler.py
jaywoo123: Yeah, I think I blindly copy and pasted. Updated.

dorisjlee:
This looks great, thanks @jaywoo123! I added some minor changes to reuse test_showMore to clean up the code. It should be ready to be merged!

PR #11, opened 2020-06-05
Mark_bar for histogram based on range of plot ratio

caitlynachen:
Before
Screen Shot 2020-06-04 at 5 04 46 PM
After
Screen Shot 2020-06-04 at 5 07 42 PM
Screen Shot 2020-06-04 at 5 06 56 PM
Screen Shot 2020-06-04 at 5 01 52 PM


dorisjlee:
Thanks @caitlynachen! I added some description to this PR and additional code to reflect this change in the exported Altair code. In the future, it would be useful include more description to the pull request for reference.

PR #12, opened 2020-06-05
Add aggregation and generalized channel output in View and View Collection __repr__

westernguy2:


dorisjlee:

PR #17, opened 2020-06-21
Add PeriodIndex compatibility and Datetime demo notebook

westernguy2:
Allows users to set their desired granularity for any datetime-like columns using PeriodIndex.


dorisjlee:
This looks good @westernguy2, I made some changes to the demo file to add to the narrative. However, I noticed that the dataType of the `date` field is not being recognized as a temporal field even after we do the `to_datetime` (See In[9] and [10] in screenshot and notebook). Could you look into how we can fix this? Thanks!
image

dorisjlee:
Merging in this commit, we will defer the bugfix for missed call to `computeDatasetMetadata` to a later PR.

PR #19, opened 2020-06-23
Display input for view and view collection in __repr__ even when data not yet loaded

westernguy2:


dorisjlee:
Looks great, the rebasing did simplify the commit history. Thanks @westernguy2 !

PR #20, opened 2020-06-26
Extend __setitem__ to ensure metadata gets refreshed for inplace changes

westernguy2:
Also added a test that tests this issue, and fixed output for DatetimeDemo notebook.


westernguy2:
I also found a bug with the old View repr, so I fixed that

dorisjlee:
Looks great, thanks @westernguy2 ! This PR and #17  means that Lux have a better support for datetime fields.

PR #21, opened 2020-07-05
Interestingness Bug Fix

19thyneb:
Fixed issue where view with single point was top recommended graph for Scatter Plot filter action.


dorisjlee:
This looks good, thanks @thyneb19 !

PR #22, opened 2020-07-07
Updated Barchart Interestingness Function

19thyneb:
Added rank based significance factor to filtered barchart scoring.


lux/interestingness/interestingness.py
```diff
@@ -127,8 +127,25 @@ def deviationFromOverall(view:View,ldf:LuxDataFrame,filterSpecs:list,msrAttribut
...
 	sig = v_filter_size/v_size #significance factor
 	# Euclidean distance as L2 function
+
+	rankSig = 1 #category measure value ranking significance factor
+	#if the view is a barchart, count how many categories' rank, based on measure value, changes after the filter is applied
+	if view.mark == "bar":
+		dimList = view.getAttrByDataModel("dimension")
+
+		#use Pandas rank function to calculate rank positions for each category
+		v_rank = unfilteredView.data.toPandas().rank().toPandas()
```
dorisjlee, line 137: Is there a specific reason why we need to do `toPandas` here? Can we not apply `unfilteredView.data.rank()` directly, since LuxDataFrame inherits from Pandas?

dorisjlee:
This approach is quite clever! When we initially discussed this, I thought this was a replacement to the deviation based metric, but the approach that you've taken where the rank information is a factor multiplied to the rest of the calculation intuitively combines the benefits of the two methods. 

It would be useful to include some interestingness test to evaluate the relative order of the views, maybe by starting with your `Filter` example that motivated this change.

lux/interestingness/interestingness.py
19thyneb: Yeah we can just apply .rank() directly, missed that there.

dorisjlee: That sounds good, you can just make additional commits and push to your branch after you've removed the toPandas. It should automatically reflect on this pull request.

dorisjlee:
This looks really great, thanks @19thyneb !

PR #25, opened 2020-07-09
Change syntax to PEP 8 style

westernguy2:


dorisjlee:
Thanks @westernguy2 !

PR #32, opened 2020-07-17
Update to SQL Executor and SQL Executor Demo

19thyneb:
Fixed bug in SQL Executor where filtered bar charts were not being padded with empty categories.

Updated Flights SQL Executor demo. Added script that can be used to migrate flights.csv data to a postgresql database table.

Updated VisCollection to use the dataframe's executor instead of just the Pandas executor.


dorisjlee:
This looks great, thanks @19thyneb !

PR #38, opened 2020-07-19
Generalize Action Bug Fix, Copy VisSpec List bugfix

19thyneb:
Issue where copies of LuxDataFrame contexts were not resulting in a list of brand new VisSpecs (resulted in list of references to original VisSpecs). This resulted in issues in the compiler's determine_encoding function. When the compiler would change the encoding of one VisSpec, this also changed the encoding other copies of the VisSpec being used elsewhere.

Bug was happening when more than 2 variables were being specified in a LuxDataFrame's context such as:
`df.set_context(["Cylinders", "Horsepower", "MilesPerGal"])`

Added copy functions to create a true copy of an individual VisSpec and a function to create a copy of a LuxDataFrame's context.

### DIFF
--- a/lux/luxDataFrame/LuxDataframe.py
+++ b/lux/luxDataFrame/LuxDataframe.py
@@ -56,7 +56,7 @@ class LuxDataFrame(pd.DataFrame):
         elif (type.lower()=="pandas"):
             self.default_pandas_view = True
         else: 
-            warnings.warn("Unsupported display type. Default display option should either be `lux` or `pandas`.",stacklevel=2)
+            warnings.warn("Unsupported display type. Default display option should either be `lux` or `pandas`.")
     # @property
     # def context(self):
     #     return self.context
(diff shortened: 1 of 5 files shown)


dorisjlee:
This looks good. Can you add a test case for this? 
Also can we replace what's inside `copy_spec` as a shallow copy of the object properties via __dict__? This will make the copy_spec to be more easily maintainable if we decide to add additional attributes to `VisSpec`.
https://stackoverflow.com/questions/243836/how-to-copy-all-properties-of-an-object-to-another-object-in-python

19thyneb:
I can add a test case for this. From what I found the bug seemed to only be isolated within the generalize action and caused problems in the compiler, but I did not have to make any changes to the compiler for the fix.

Will look into the shallow copy solution to make the copy function easier to maintain.

dorisjlee:
Closed, merged in #42 instead.

PR #42, opened 2020-07-25
Remove Column and Filter from spec bug fix

19thyneb:
Fixed bug where remove_column_from_spec and remove_filter_from_spec functions were not properly removing Clauses from Vis's query and inferred query. 
This fixed a bug in the generalize action where the original graph was being displayed multiple times.


dorisjlee:
Thanks @thyneb19 !

PR #46, opened 2020-07-26
Added Colored Barchart and Colored Linechart Functionality

19thyneb:
Executor now properly executes data for colored bar and line charts. Interestingness function now also assigns a default interestingness score to these charts as a placeholder. Additional tests were added to test_executor.py to check if the pandas executor is correctly handling colored bar and line charts.

Added test for generalize action to make sure all recommended graphs produced are distinct.


dorisjlee:
This looks great, thanks @thyneb19 !
image
image

PR #47, opened 2020-07-27
Pandas Executor performance improvements

jrdzha:
20+ minutes down to ~1 second on 20K Rows of Amazon Reviews Dataset.


dorisjlee:
Thanks @jrdzha ! I tested the performance of this on the census dataset (4MB), and it went from 14 seconds to around 5 seconds.

Issue #53, opened 2020-08-04
Sphinx Autodoc doesn't create correct hyperlinks for modules


westernguy2:
Any Sphinx autogenerated links to Lux modules gets hyperlinked to the bottom of the page.

For example:

At [this link](https://lux-api.readthedocs.io/en/latest/py-modindex.html), if you press `lux.action`, you get directed to the bottom of the page, specifically [here](https://lux-api.readthedocs.io/en/latest/source/reference/lux.action.html#module-lux.action).

For some reason, all links to modules go directly to the "Module contents" section of a page, which happens to be at the bottom.

This issue was brought up at [this StackOverflow post](https://stackoverflow.com/questions/35810213/can-sphinx-emit-the-module-contents-first-and-the-submodules-last), but the fix is for the sphinx library command rather than something we can add to `conf.py` or one of the rst files. However, that may be a good place to start when looking for a fix.

dorisjlee:
The module is linking to the right places now, but the packages still bring us to the bottom of the page. This is okay since it's still bringing us to the right page. Also with the API page working, this is no longer an issue.

PR #58, opened 2020-08-07
Add API page to Documentation

westernguy2:
The issue at #53 isn't necessary to fix anymore since the API now links directly to the classes instead of the modules.


dorisjlee:
Thanks @westernguy2, this looks great !

Issue #52, opened 2020-07-30
Interestingness Ranking of colored v.s. no colored bars


dorisjlee:
Colored bars and lines currently have a fixed score of 0.2, which is perhaps a bit too high. There might be an issue with the interestingness calculation of the no colored bars since their values are all close to zero. Ideally, the no colored bars should be prioritized since they are simpler and more interpretable.
![image](https://user-images.githubusercontent.com/5554675/88913791-34bcc280-d294-11ea-8919-d97e8c1bc381.png)
![image](https://user-images.githubusercontent.com/5554675/88913830-430ade80-d294-11ea-9e92-ff9b72a5bc14.png)

dorisjlee:
With the new PR pulled in, the interestingness score for the colored visualizations are now dynamic. But the non-colored vis is still not showing up in the beginning. The visualization with Name should still be ranked lower than the rest. There might be something weird happening with the `MEAN(*), Cylinders` visualization since their scores are all close to 0.

![image](https://user-images.githubusercontent.com/5554675/90262827-654e5000-de81-11ea-80e7-e112bac2836c.png)

dorisjlee:
This is fixed by @thyneb19  in #72. The `Name` visualization is deprioritized due to high cardinality. The cardinality downweighting factor is decreased in the unevenness metric.

![image](https://user-images.githubusercontent.com/5554675/90990392-aa862680-e5d3-11ea-8c67-5f9315eed4a8.png)
![image](https://user-images.githubusercontent.com/5554675/90990399-b245cb00-e5d3-11ea-9260-10d0bfa4fbe6.png)

PR #59, opened 2020-08-09
Added Interestingness Scoring for Colored Bar and Line charts

19thyneb:
Added test case to test_interestingness.py under test_interestingness_1_2_0 function. For addressing issue #52


19thyneb:
Fixed bug in the Pandas executor where colored bar/line chart meta data was not being calculated. This also decreased performance on the census data by around 1.5 seconds.

dorisjlee:
This seems to be causing an issue in `test_q1_performance_census`? Not sure why its doing this.

19thyneb:
@dorisjlee so the error in the build check was due to the metadata of certain Vis objects not being calculated. So in that case, the vis.data.cardinality object was of NoneType, which caused the key error.

I've fixed this issue and the test runs fine locally, not sure why the build isn't working properly after I pushed my changes though.

lux/executor/PandasExecutor.py
```diff
@@ -200,6 +200,11 @@ def execute_aggregate(view: Vis,isFiltered = True):
                     for col in columns[1:]:
                         view.data[col] = view.data[col].fillna(0)
                     assert len(list(view.data[groupby_attr.attribute])) == len(all_attr_vals), f"Aggregated data missing values compared to original range of values of `{groupby_attr.attribute}`."
+            #need to compute the statistics and metadata for the view's data if no new rows were added
+            else:
+                if view.data.cardinality is None and has_color:
```
dorisjlee, line 205: Is there a reason why we need to recompute the metadata here?

dorisjlee:
Merging this in for now, will see if the new metadata refresh changes will fix this.

dorisjlee:
This issue is fixed after we pull in the latest code for metadata maintenance. I tried to remove the hack line 203-205 in PandasExecutor.py but it is still required. This is happening because when we do 
`view.data = groupby_result.agg(agg_func).reset_index()`
The metadata is not propagated over in `__finalize__`. In particular, the metadata is lost in the step where we perform the aggregation `groupby_result.agg(agg_func)`.
This is happening because `groupby_result` is a [GroupBy object](https://pandas.pydata.org/pandas-docs/stable/reference/groupby.html). This means that when we are propagating the metadata over, it is lost when the DataFrame gets transformed into the GroupBy object.
I resolved this bug in commit [eaed32a](https://github.com/lux-org/lux/commit/eaed32a9622d5d91db51a34e5ebe071daae9b732) by explicitly passing the original metadata to the resulting intermediate result using `__finalize__`.
```
groupby_result = groupby_result.agg(agg_func)
intermediate = groupby_result.reset_index()
view.data = intermediate.__finalize__(view.data)
```

Issue #63, opened 2020-08-14
Date Visualization Bug in Flights Dataset


dorisjlee:
X axis for temporal attribute looks off for the flights dataset.
![image](https://user-images.githubusercontent.com/5554675/90230964-f8b95e00-de4c-11ea-9181-0e80801b2d69.png)

dorisjlee:
This issue is resolved if we add `pd.to_datetime` statements, automatically inferred temporal attributes are not automatically converted to datetime.
```
df = pd.read_csv("lux/data/flights.csv")
df["year"] = pd.to_datetime(df["year"], format='%Y') 
df["month"] = pd.to_datetime(df["month"], format='%M')
df["day"] = pd.to_datetime(df["day"], format='%d')
df
```
![image](https://user-images.githubusercontent.com/5554675/91117389-0bd7f380-e6c1-11ea-84c0-8dd3000f7b6c.png)

dorisjlee:
This issue is resolved by adding a warning message when Lux suspects that a data type is temporal but the Pandas type is not a datetime object. We thought about ways of performing automatic formatting in `to_datetime`, however, with the detection of days and months would require more careful and brittle heuristics. We opted for showing the warning and let the user specify their own `pd.to_datetime` statements and format.

PR #80, opened 2020-08-30
Add warning when temporal column doesn't have datetime type

westernguy2:
When Lux infers that a column is temporal, a warning is thrown if that column is not of type datetime. Attached is a picture of the warning.

Picture of Warning

This should fix the issue in #63 

Also fixed a bug in `__setitem__` that was causing metadata to be computed multiple times at each `_repr_html_` call.


westernguy2:
Reverted the fix to `__setitem__` because it was causing tests to fail. Issue detailed in #81 . Also fixed issue where `pd.Timestamp` objects were not being added correctly to the `data_type_lookup` dictionary.

dorisjlee:

PR #87, opened 2020-09-18
Update export tutorial to add explanation for to_Altair's standalone argument

westernguy2:
Updated both rst file and tutorial notebook.


dorisjlee:
Thanks @westernguy2! I made minor modifications to the text and cleaned up the ipynb before saving. We generally click on "Kernel --> Restart & Clear Output" in the Jupyter notebook and save it in this "clean" state, so that the notebook doesn't contain the output history, when you do this you should see that `"execution_count": null` in each of the cells in the raw ipynb file.

PR #90, opened 2020-09-19
Moved Metadata functions to Executors, resolved issue in test_prefix case

thyneb19:
Fixed test case test_prefix in pandas_coverage function, 1_year variable that was created was not getting it's data type computed


dorisjlee:
This is a much cleaner division between the Executor and frame.py, thanks @thyneb19 !

Issue #79, opened 2020-08-27
Unsupported Category dtype as_type


dorisjlee:
There is a bug when using Lux with [this example](https://datashader.org/getting_started/Pipeline.html) from Datashader.

```python
import pandas as pd
import numpy as np
from collections import OrderedDict as odict

num=10000
np.random.seed(1)

dists = {cat: pd.DataFrame(odict([('x',np.random.normal(x,s,num)), 
                                  ('y',np.random.normal(y,s,num)), 
                                  ('val',val), 
                                  ('cat',cat)]))      
         for x,  y,  s,  val, cat in 
         [(  2,  2, 0.03, 10, "d1"), 
          (  2, -2, 0.10, 20, "d2"), 
          ( -2, -2, 0.50, 30, "d3"), 
          ( -2,  2, 1.00, 40, "d4"), 
          (  0,  0, 3.00, 50, "d5")] }

df = pd.concat(dists,ignore_index=True)
df["cat"]=df["cat"].astype("category") #  If commented, the df.intent=vis line doesn't break

df  #Select and export the scatterplot vis (big circular blob)

vis = df.exported[0]

df.intent = vis 
df # This errors on the key 'cat' which is a categorical
```
![image](https://user-images.githubusercontent.com/5554675/91453404-405ed180-e8b2-11ea-8722-2b817d75f754.png)

We should look into supporting [Categorical](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html) data types in Pandas. More importantly, we should look into whether bugs show up when we perform `astype` operations for other data types.

westernguy2:
This is another bug where Lux isn't supporting CategoricalDtype. Specifically, both `cut` and `qcut` output a column of CategoricalDtype. 

```python
import pandas as pd
import numpy as np
import lux

url = 'https://github.com/lux-org/lux-datasets/blob/master/data/cars.csv?raw=true'
df = pd.read_csv(url)
df['Year'] = pd.to_datetime(df['Year'], format='%Y')

df["Weight"] = pd.qcut(df["Weight"], q = 3)

```

```python
import pandas as pd
import numpy as np
import lux

url = 'https://github.com/lux-org/lux-datasets/blob/master/data/cars.csv?raw=true'
df = pd.read_csv(url)
df['Year'] = pd.to_datetime(df['Year'], format='%Y')

df["Weight"] = pd.cut(df["Weight"], bins = [0, 2500, 7500, 10000], labels = ["small", "medium", "large"])

[...]

Issue #66, opened 2020-08-18
Better support for Pandas.Series


dorisjlee:
We should create a LuxSeries object to take on the sliced version of the LuxDataframe, following [guidelines](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#override-constructor-properties) for subclassing DataFrames. We need to pass the `_metadata` from LuxDataFrame to LuxSeries so that it is preserved across operations (and therefore doesn't need to be recomputed), related to #65. Currently, this code is commented since LuxSeries is causing issues compared to the original Pd.Series. 

```python
class LuxDataFrame(pd.DataFrame):
        ....
	@property
	def _constructor(self):
		return LuxDataFrame

	@property
	def _constructor_sliced(self):
		def f(*args, **kwargs):
			# adapted from https://github.com/pandas-dev/pandas/issues/13208#issuecomment-326556232
			return LuxSeries(*args, **kwargs).__finalize__(self, method='inherit')
		return f
```

```python
class LuxSeries(pd.Series):
	# _metadata =  ['name','_intent','data_type_lookup','data_type',
	# 			 'data_model_lookup','data_model','unique_values','cardinality',
	# 			'min_max','plot_config', '_current_vis','_widget', '_recommendation']
	def __init__(self,*args, **kw):
		super(LuxSeries, self).__init__(*args, **kw)
	@property
	def _constructor(self):
		return LuxSeries

	@property
	def _constructor_expanddim(self):
		from lux.core.frame import LuxDataFrame
		# def f(*args, **kwargs):
		# 	# adapted from https://github.com/pandas-dev/pandas/issues/13208#issuecomment-326556232
		# 	return LuxDataFrame(*args, **kwargs).__finalize__(self, method='inherit')
		# return f
		return LuxDataFrame
```
In particular the original `name` property of the Lux Series is lost when we implement LuxSeries, see `test_pandas.py:test_df_to_series` for more details.
Example: 
```python
df = pd.read_csv("lux/data/car.csv")
df._repr_html_()
series = df["Weight"]
series.name # returns None (BUG!)
series.cardinality # preserved
```
We should also add a __repr__ to print out the basic histogram for Series objects.

dorisjlee:
Support for Pandas Series is much more robust now with these changes. Thanks @westernguy2 !

PR #102, opened 2020-09-30
Adding support for all dtypes, adds more tests and fixes small bugs

westernguy2:
This pull request solves the issue in #79 where the dtype of a column is not being assigned a data_type in `data_type_lookup`, resulting in an error. This is resolved by setting any column that has a dtype not being accounted for as `nominal`. In addition, it became clear that the `ordinal` data type wasn't being used, so that was removed from Lux.

Also fixed are bugs relating to `__getattr__` due to a missing return value, which fixed the issue where accessing columns of a DataFrame via dot notation wasn't working. 

Finally, some more tests for `.loc` and `.iloc` were added, as well as a new test for `groupby.agg` which hasn't been resolved yet (in issue #66 ).


dorisjlee:
This looks great! One remaining issue in #79 is the use of `qcut`, which creates Interval data type, it seems like there is some difficulty in parsing this into the widget representation (i.e. JSON serializable). The use of `cut` seems to be working fine. I've added a test for this in the PR. 
image

westernguy2:
This commit checks if dtype of a column is of type `Interval` in `create_vis`. If it is, then the column is converted to type `str` instead. This only modifies the visualization's data, not the original LuxDataFrame. This should fix the issue where columns of type `Interval` are unable to be visualized, as now it's a String type.

In addition, metadata is also expired in `_set_item` as there was a bug where direct column assignment was not updating the metadata accurately.

dorisjlee:

PR #105, opened 2020-10-03
Added color for bar charts

### LINKED ISSUE #67: Assign unique color to different values in unsorted bar charts
Currently, all bars in the bar chart has the same color, this makes comparison across items challenging. In particular, users need to read the bar labels to realize that the bars are sorted to match other charts for comparison. We should assign a consistent set of colors for low cardinality categorical attributes to facilitate ease of comparison. We can use a standard color pallette for the particular visualization library ([Vega](https://vega.github.io/vega/docs/schemes/#categorical), [matplotlib](https://matplotlib.org/3.1.1/tutorials/colors/colormaps.html)).

image
The update bars would look something like this, where all values of Cylinder=4 is green, all values of Cylinders of 8 is yellow, etc.
image

jerrysong1324:
Needed to use if clause to check that the chart did not already have color attributes assigned. If it did, adding color based on Y attribute would override that and cause confusion. Fixes #67


jerrysong1324:
Video Demo: https://www.loom.com/share/ef7d3cfa036b4dd9b128c67d609e9205

dorisjlee:
This looks like a good start!  It would be useful to have a separation between the *basic* bar charts where the color is the same as the x attribute (i.e., the bars that didn't use to have color) with the *enhanced* bar charts that actually had color associated with it in the first place. There are several things that we can do to separate these: 
1) Change the color palette of the *basic* bar chart so that it is distinctly different from the *enhanced* bar charts. We should make sure that there is no overlap between the two chosen color scheme to avoid confusion (see the [categorical color schemes in Vega](https://vega.github.io/vega/docs/schemes/)). In particular, the bars should ensure that *across all charts* all values of Cylinder=4 is green, all values of Cylinders of 8 is yellow, etc. My guess is that a generic color assignment via Altair like what we currently have wouldn't work here, we would need something more global to keep track of things across visualizations so that we know what values correspond to what color. 

2) Remove color legend on the *basic* bar chart, since this provides redundant information as what's shown in the axes. This can be done via: `alt.Color('COLOR_ATTR',legend=None)`

jerrysong1324:
@dorisjlee This second point is addressed in the above commit. The first point is a bit trickier. If I'm understanding correctly, we want these color to persist across all bar charts (basic and enhanced) as well as all other plots potentially as well (line, scatter, etc.)?

dorisjlee:
Yeah, we should persist the same color across all bar charts (basic and enhanced). I hadn't really thought about extending this to other plots, although that would be nice. For now, I think we can focus on making sure there is a unique color associated with each bar value.

dorisjlee:
Closing this for now.

Issue #89, opened 2020-09-18
Avoid showing histograms for low cardinality quantitative attributes


dorisjlee:
There is a bug related to quantitative attributes that have low cardinality. For example, we create a column of values where all the values are `4.0`. There is a warning message resulting from a divide-by-zero error and a histogram that looks, since the range of the computed min/max field is zero. This issue will be resolved if we avoid computing histograms for quantitative values that have low cardinality.

```
import pandas as pd
import lux
df = pd.read_csv("../../lux/data/car.csv")
df["Year"] = pd.to_datetime(df["Year"], format='%Y') # change pandas dtype for the column "Year" to datetype
df["Units"] = 4.0
df
```

![image](https://user-images.githubusercontent.com/5554675/93612168-c11c7380-fa01-11ea-9253-d03b4f192168.png)

PR #106, opened 2020-10-06
Remove histograms with low-medium cardinality.

jinimukh:
This PR addresses Issue 89 and stops histograms for low-cardinality columns to be computed/suggested. Another divide-by-zero error is caused by correlation of low-cardinality columns -- this will be fixed in a later PR.

As part of the implementation, I have commented the optimization on computing cardinality for columns of type float64. 

Screen Shot 2020-10-05 at 5 51 43 PM


dorisjlee:
This PR resolves #89. The performance test is passing after merging into master.
Thanks @jinimukh!

PR #112, opened 2020-10-14
Adding support for setting intent on front end

cjachekang:
Front-end PR: https://github.com/lux-org/lux-widget/pull/48

-Added support to the front end by setting intent in the backend and then destroying and re-initializing widget.


lux/core/frame.py
```diff
@@ -537,6 +537,25 @@ def removeDeletedRecs(self, change):
 				self.recommendation[action].remove_index(index - deletedSoFar)
 				deletedSoFar += 1
 
+	def setIntentOnButtonClick(self, change):
```
dorisjlee, line 540: Function names should be snake_case based on PEP8 style.

lux/core/frame.py
```diff
@@ -537,6 +537,25 @@ def removeDeletedRecs(self, change):
 				self.recommendation[action].remove_index(index - deletedSoFar)
```
dorisjlee, line 537: Also change removeDeletedRecs as snake_case based on PEP8 style.

dorisjlee:
We should update the tutorial and readthedoc page with information about setting intent as vis and how to do that via the UI or code. We currently describe how to do this in code in [this notebook](https://github.com/lux-org/lux-binder/blob/master/tutorial/3-widget-vis-export.ipynb) but we could do it wherever it seems most appropriate. The tutorial can be a separate PR.

dorisjlee (approved):

Issue #108, opened 2020-10-06
High-cardinality bar chart text color overridden by color channel encoding


dorisjlee:
For high cardinality bar chart, we have a text indicator that shows "X more..." bars, however the text attribute color:
```
text = alt.Chart(visData).mark_text(
	x=155, 
	y=142,
	align="right",
	color = "#ff8e04",
	fontSize = 11,
	text=f"+ 230 more ..."
)
```
is overridden by the color encoding:
```
chart = chart.encode(color=alt.Color('occupation',type='nominal'))
```
which is generated from `AltairChart:encode_color`, the encode color is called last since that is only added for visualizations that have a color attribute.

This example can be found in the census dataset.
```
df = pd.read_csv("../lux-datasets/data/census.csv")
df.set_intent(["education"])
df
```
![image](https://user-images.githubusercontent.com/5554675/95207368-5f8a4080-081a-11eb-9a1f-0050ea0b3a62.png)
![image](https://user-images.githubusercontent.com/5554675/95207407-6a44d580-081a-11eb-9b75-492c2d05417e.png)
![image](https://user-images.githubusercontent.com/5554675/95207498-86487700-081a-11eb-81ed-1782e67145d2.png)

PR #115, opened 2020-10-16
Encode color of chart before adding text

jinimukh:
This PR solves #108 where the text attribute was getting overridden by the encoding color. Since encoding takes priority over any attributes we set on the chart, the fix, as recommended by https://github.com/altair-viz/altair/issues/1033 is to encode the chart separately from the text. I overrode the `encode_color()` method in `BarChart.py` by adding the text after the encoding. All other graphs behave the same way. 

Screen Shot 2020-10-18 at 10 37 13 PM

### DIFF
--- a/examples/demo/college_demo.ipynb
+++ b/examples/demo/college_demo.ipynb
@@ -121,7 +121,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "df.set_intent([\"AverageCost\"])\n",
+    "df.intent = [\"AverageCost\"]\n",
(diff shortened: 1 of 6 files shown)


dorisjlee:
We need to make sure that `encode_color` still works for the other chart types. So we should still have the global `encode_color`, but instead triggering it within each chart before any text specification, instead of after `initialize_chart`.

dorisjlee:
Tested that this is working and made minor fixes. Thanks @jinimukh !

Issue #113, opened 2020-10-16
No module named 'luxwidget' when installing and activating the api


Smartog:
The console leaves me a "ModuleNotFoundError: No module named 'luxwidget'" when i try to run "jupyter nbextension install --sys-prefix --symlink --overwrite --py luxwidget" in cmd.

dorisjlee:
Hi @Smartog,

You can follow the quick install instructions [here](https://github.com/lux-org/lux#quick-installation).
When you do `pip freeze` do you see `lux-widget` listed under one of the installations? 
`lux-widget` should have been installed when you did `pip install lux-api` as the first installation step. 
If not, you may want to do `pip install lux-widget` to install it directly. 
Please let me know if you are still experiencing issues.

Smartog:
Yeal,It works! @dorisjlee ,
After running the "pip freeze", I found the "luxWidget==0.0.7" and  the letter "w" is capital that is why I failed when spelling it "luxwidget".

Yet,another problem occurred when I tried to run:
`import lux
import pandas as pd
df = pd.read_csv("college.csv")
df
`

There should be a buttom to show the picture below but i cannot find it. 
The error indication is:
`
d:\python3.8.6\lib\site-packages\IPython\core\formatters.py:345: UserWarning: 
Unexpected error in rendering Lux widget and recommendations. Falling back to Pandas display.

Please report this issue on Github: https://github.com/lux-org/lux/issues 
  return method()
`

Smartog:
@dorisjlee 
sorry,the code I run is:
`import lux`
`import pandas as pd`
`df = pd.read_csv("college.csv")`
`df`
I am a freshman in the github editor.

dorisjlee:
Hi @Smartog, 

It seems like you have an older version of luxwidget (in v0.0.7, we had the capital luxWidget, but we have changed it in the latest version). So it's likely that your luxwidget and lux-api versions are not in sync. I would encourage you to install both your lux-api and lux-widget:
```
pip uninstall lux-api
pip uninstall lux-widget

jupyter nbextension uninstall --py luxWidget
jupyter nbextension disable --py luxWidget
```
Then try doing the install again (the new lux-api should just install the widget for you directly): 
```
pip install lux-api

 jupyter nbextension install --py luxwidget
 jupyter nbextension enable --py luxwidget
```

Smartog:
Thank you !!!@dorisjlee
[...]

PR #117, opened 2020-10-19
Update FAQ

jinimukh:
Updated FAQ with information from https://github.com/lux-org/lux/issues/113 and virtual environment fix


dorisjlee:
Thanks @jinimukh !

Issue #65, opened 2020-08-14
Metadata propogation not reliable if object type changes


dorisjlee:
We currently rely on Pandas's [__finalize__](https://github.com/pandas-dev/pandas/blob/master/pandas/core/generic.py#L5179) to propagate the _metadata (including computed metadata, recommendations, and other stored properties). However, there are situations where dataframes don't always stay the same type throughout various operations. For example, in this [issue](https://github.com/lux-org/lux/pull/59#issuecomment-674100795), when we do group by. We end up with a GroupBy object, and the `_metadata` is lost. Similar issues will occur when going from Pandas Dataframe --> Series. 
We should find a better strategy for metadata maintenance without having to explicitly pass in __finalize__, or ensure that when metadata properties are retrieved that they trigger a metadata recomputation if unfresh (so that the system is slightly slower but doesn't break when the metadata fields are accessed).

```python
df = pd.read_csv("lux/data/car.csv")
groupby_result = df.groupby("Origin").agg("sum")
intermediate = groupby_result.reset_index()
intermediate.cardinality # this returns None
df = intermediate.__finalize__(df)
df.cardinality # cardinality is now populated again
```

westernguy2:
There also seems to be an issue of metadata propagation with `.loc` and `.iloc`. For example, `df.loc` returns a `LocIndexer` object, which we can index into to get/set parts of the DataFrame. However, since we have an intermediate object, the metadata is lost.

This is seen in the CollegeDemo notebook:

```python
df = pd.read_csv("../../lux/data/college.csv")
df = df.drop(columns=["ACTMedian"])
df.loc[df["PredominantDegree"]=="Certificate","PredominantDegree"] = "Associate"
df.unique_values()["PredominantDegree"]
```

This code replaces all instances of "Certificate" in the `PredominantDegree` column with "Associate", and so the unique values in that column will no longer contain "Certificate", however it still does. Clearly, the metadata is not being expired and recomputed.

Issue #97, opened 2020-09-25
HTTP Error 429: too many requests causes tests to fail


dorisjlee:
Occasionally, the test suite can fail when it fetches the data from the same URL too frequently, we should replace the URL link with local versions of the cars.csv if possible. In addition, any remote fetches of the data should catch the HTTP 429 error and add a sleeping timer to ensure that the test doesn't fail when the query rate is too high.
We could also look into dataframe reuse opportunities across tests so that the dataframe doesn't need to be reloaded every single time.
![image](https://user-images.githubusercontent.com/5554675/94279144-4a8df180-ff7e-11ea-86c5-e4e803a875be.png)

PR #122, opened 2020-10-25
Add LuxSeries Implementation

westernguy2:
Adds implementation for LuxSeries inline with [Pandas guidelines](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures) on extending `DataFrame` and `Series`.

This will fix most issues where metadata isn't being propagated across different method calls, specifically when there is an intermediate object (that's not a DataFrame) such as a Series object or a GroupBy object. The issue in #66 has been fixed, and the first example in #65 has also been fixed.

This pull request also adds minor bug fixes and adds tests for the LuxSeries implementation.


lux/core/series.py
```diff
@@ -14,20 +14,26 @@
...
 	def __init__(self,*args, **kw):
 		super(LuxSeries, self).__init__(*args, **kw)
+
 	@property
 	def _constructor(self):
 		return LuxSeries
 
 	@property
 	def _constructor_expanddim(self):
 		from lux.core.frame import LuxDataFrame
```
dorisjlee, line 29: Is there a reason why we don't need to do __finalize__ here? I tested out the example [here](https://github.com/lux-org/lux/issues/65#issue-679200159) but the metadata did not propagate without the explicit `__finalize__`.

lux/core/series.py
```diff
@@ -14,20 +14,26 @@
...
-		# def f(*args, **kwargs):
```
dorisjlee, line 29: Is there a reason why we don't need to do `__finalize__` here? I tested out the example [here](https://github.com/lux-org/lux/issues/65#issue-679200159) but the metadata did not propagate without the explicit `__finalize__`.

dorisjlee (changes requested):

lux/core/series.py
```diff
@@ -14,20 +14,26 @@
...
-		# 	# adapted from https://github.com/pandas-dev/pandas/issues/13208#issuecomment-326556232
-		# 	return LuxDataFrame(*args, **kwargs).__finalize__(self, method='inherit')
-		# return f
+	# 	def f(*args, **kwargs):
```
jrdzha, line 30: Is there a reason that we need to leave this commented? If so, it would be good to have another comment explaining why above this section. Or a TODO tag if deferring to a later date.

tests/test_pandas_coverage.py
```diff
@@ -355,39 +355,39 @@ def compare_vis(vis1, vis2):
...
+    assert df.cardinality is not None
+    series = df["Weight"]
+    series.value_counts()
+    assert isinstance(series,lux.core.series.LuxSeries), "Derived series is type LuxSeries."
+    assert df["Weight"]._metadata == ['_intent','data_type_lookup','data_type','data_model_lookup','data_model','unique_values','cardinality','_rec_info','_pandas_only','_min_max','plot_config','_current_vis','_widget','_recommendation','_prev','_history','_saved_export'], "Metadata is lost when going from Dataframe to Series."
+    assert df.cardinality is not None, "Metadata is lost when going from Dataframe to Series."
+    assert series.name == "Weight", "Pandas Series original `name` property not retained."
+
+def test_str_replace():
+    url = 'https://github.com/lux-org/lux-datasets/blob/master/data/cars.csv?raw=true'
```
dorisjlee, line 384: replace read_csv url to local reference of `car.csv` to avoid #97

lux/core/series.py
westernguy2: I believe that `__finalize__` is done implicitly when we call `_repr_html_`. The example given works if we call `_repr_html_` before querying for the cardinality.

tests/test_pandas_coverage.py
westernguy2: Fixed, except for the last test case which uses the updated `cars` dataset via the url. Should I rewrite the test and use the local version?

lux/core/series.py
westernguy2: Removed the comments, since we won't be needing to using `__finalize__` anymore.

dorisjlee (approved):

dorisjlee:
I've added a few bugfix to resolve to create a bar chart with respect to the group by attribute, by iterating only over the quantitative attributes in the column group action. Another fix was that I simplified and consolidated the logic inside column group so that bar charts are displayed for `test_metadata_column_group_reset_df`.

PR #125, opened 2020-10-26
Enhanced datetime detection added to compiler

jrdzha:
Adding enhanced datetime detection (automatic detect string types).

First part in implementing automatic datetime feature extraction.


lux/processor/Compiler.py
```diff
@@ -218,13 +218,15 @@ def determine_encoding(ldf: LuxDataFrame, vis: Vis):
 		nmsr = 0
 		filters = []
 		for clause in vis._inferred_intent:
+			# print('clause', clause)
```
dorisjlee, line 221: Remove prints and whitespace

lux/executor/PandasExecutor.py
```diff
@@ -317,14 +318,20 @@ def compute_dataset_metadata(self, ldf:LuxDataFrame):
...
         for attr in list(ldf.columns):
-            temporal_var_list = ["month", "year","day","date","time"]
-            if (isinstance(attr,pd._libs.tslibs.timestamps.Timestamp)): 
+            if is_datetime(ldf[attr]):
+                ldf.data_type_lookup[attr] = "temporal"
+            elif (isinstance(attr,pd._libs.tslibs.timestamps.Timestamp)): 
                 # If timestamp, make the dictionary keys the _repr_ (e.g., TimeStamp('2020-04-05 00.000')--> '2020-04-05')
                 ldf.data_type_lookup[attr] = "temporal"
             # elif any(var in str(attr).lower() for var in temporal_var_list):
             elif str(attr).lower() in temporal_var_list:
```
dorisjlee, line 331: should we get rid of this check now that we have the more robust datetime checker?

lux/core/frame.py
```diff
@@ -609,7 +611,9 @@ def on_button_clicked(b):
 					
 		except(KeyboardInterrupt,SystemExit):
 			raise
-		except:
+		except Exception as e:
+			warnings.warn(e)
+			warnings.warn(traceback.format_exc())
```
dorisjlee, line 616: What does the `format_exc` do here?

dorisjlee (changes requested):

dorisjlee:
Thanks @jrdzha! I've added some comments via code review. Could you add some test cases and examples to show that this is working?

lux/executor/PandasExecutor.py
```diff
@@ -317,14 +318,20 @@ def compute_dataset_metadata(self, ldf:LuxDataFrame):
         self.compute_data_model(ldf)
 
     def compute_data_type(self, ldf:LuxDataFrame):
+        from pandas.api.types import is_datetime64_any_dtype as is_datetime
+        temporal_var_list = ["month", "year","day","date","time"]
```
thyneb19, line 322: For the temporal variable name check, should we include other potential temporal variables to the list? Eg: hour, minute, second, etc.

jrdzha: This was code that was there already, so I left it as is. I think it's reasonable to add the others as well. At the same time, I'm not sure that reading column names to detect temporal is good practice to begin with. For example, one of my datasets had "review_month" as a column. That was just recognized as an integer column. This will also fail on columns that are titled "month", but contain string values (ex. January, February, etc.). Thoughts?

@dorisjlee thoughts?

lux/core/frame.py
jrdzha: It just gives the full stack trace rather than just saying that lux-widget failed. This is just more robust issue reporting so if users encounter this, they know what to post on a Github issue. It just makes this show up as a more normal python error while also letting the user continue using the pandas view. I doubt that this will be used often if we put out compatible lux and lux-widget versions and stay up to date on requirements.txt, but in the case where it does fail, this will give much more insight into why.

lux/executor/PandasExecutor.py
jrdzha: Line 333 was the more robust datetime checker, but that is for detecting string columns that should be datetime. This line was for detecting columns that the users have already casted to datetime, so this line would still be necessary.

lux/core/frame.py
dorisjlee: Ok that makes sense. I've modified this a bit so that the `warnings.warn(e)` doesn't throw an error.

lux/executor/PandasExecutor.py
dorisjlee: Can we actually remove this code altogether?
```
 elif str(attr).lower() in temporal_var_list:
                ldf.data_type_lookup[attr] = "temporal"
```
and rely on `is_datetime_string` to do the automatic detection? 
Like you mentioned, the string checking technique leads to a lot of false positives.

lux/executor/PandasExecutor.py
dorisjlee: Ok this makes sense, I think I was referring to the string check on `temporal_var_list`, see above comment.

dorisjlee:
@jrdzha: I looked through and tested the code in the PR. There are couple things I wanted to ask about: 
1) We discussed doing the type check just based on a sample of the data in the column, due to concerns over performance issues, but I didn't see this in the latest code. Do you have a sense of the performance impact of `is_datetime_string`?
2) Could we add some more tests around the date type checking? Currently, we only have a few attributes in cars and Airbnb inside `test_type`. It would be useful to include a more exhaustive test suite across date checks (e.g. across different datetime granularities) Even a mock dataset would be sufficient for simulating this would be sufficient.
3) If we drop the attribute name-based heuristic for detecting temporal types, it looks like the is_datetime_string would not be sufficient in testing things like month and year. I believe a regex-based detection would be more useful here and can be included into `is_datetime_string`.

lux/executor/PandasExecutor.py
jrdzha: I had actually removed this prior to making the pull request, but it broke the case where you have an int column named 'year'. That may actually be the desired behavior, but we'll just need to adjust test cases accordingly.

dorisjlee:
@jrdzha : There are some conflicts to be resolved as well as comments on adding additional tests in the above thread.

PR #130, opened 2020-10-31
Added example notebook, fixed variable reference

thyneb19:


examples/Postgres_Executor_Example.ipynb
```diff
@@ -0,0 +1,222 @@
+{
```
dorisjlee, line 1: For notebooks, make sure that you do "Restart & Clear Output" so that the outputs are not committed to the repo.
image

lux/core/frame.py
```diff
@@ -87,7 +87,7 @@ def maintain_metadata(self):
 	def expire_recs(self):
 		self._recs_fresh = False
 		self.recommendation = {}
-		self.current_vis = None
+		self.current_vis = []
```
dorisjlee, line 90: Is there a particular reason why we are changing this? In the master branch, we have this set as None.
https://github.com/lux-org/lux/blob/master/lux/core/frame.py#L88

thyneb19: Setting the current vis to None here was causing some issues when using the SQL executor. Also since in  the __init__() function we're initializing a dataframe to have current_vis be an empty list, I thought we should keep current_vis an empty list here too.

thyneb19: The issue occurred during the dataframe initialization when set_SQL_connection() is called. When column names were  being added to the dataframe, this triggered expire_recs() and changed current_vis from an empty list to None.

dorisjlee: Ok that makes sense.

dorisjlee (approved):

PR #131, opened 2020-11-03
Double init semi-bug fix

cjachekang:
Issue: https://github.com/lux-org/lux-widget/issues/20

Fixes the double initialization of the widget.

But now the intent button does not refresh the widget automatically. In the future, we would have to figure out how to manipulate the display of the widget since we are not using `self.output` anymore.

For now, we can just press the toggle button once to refresh widget - we can have an alert pop up after hitting the intent button, telling the user to refresh. While we do lose a feature, the speed improvements could be worth it. Although using the output widget is probably the correct way to do it. 

If we were to merge this, we would need to open a new issue regarding the intent.


dorisjlee:
The PR fixes the frontend `Host is not attached error` that we saw earlier.
image
I ran some performance test with and without this fix. There is a slight difference between the single/double init (with the single init being slightly faster), but this is not significant.

For the real estate dataset, the performance difference is more visible when number of datapoints is larger (`np.geomspace(1e4, 4e5, num=9)`).
</img>

<div>
</img>
</img>
</div>

cjachekang:
Ah I see - this makes sense since the backend computations would dominate

PR #134, opened 2020-11-05
documentation for custom actions

caitlynachen:


doc/source/guide/custom.rst
```diff
@@ -0,0 +1,114 @@
...
+  :align: center
+  :alt: Displays default and user-defined actions as a VisList.
+
+Using our new action, we can modify our display to only show countries with that reach a certain threshold of GDP and see how their G10 difference compares. 
+
+.. code-block:: python
+
+    df[df["GDPPerCapita"]>40000]
+
+.. image:: https://github.com/lux-org/lux-resources/blob/master/doc_img/custom-1.png?raw=true
```
dorisjlee, line 76: This image link is the same as above. Maybe we need to upload custom-2?

doc/source/guide/custom.rst
```diff
@@ -0,0 +1,114 @@
...
+Navigating the Action Manager
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To see what actions are defined on the Lux Action Manager, the following lines allow you to navigate lux.actions to see both default and user-defined actions.
+
+.. code-block:: python
+    
+    lux.actions.__len__()
+    lux.actions.__getactions__()
+
```
dorisjlee, line 92: Add what each of the printed output looks like

doc/source/guide/custom.rst
```diff
@@ -0,0 +1,114 @@
...
+You can also get a single action attribute by calling this function with the action's id.
```
dorisjlee, line 93: Do we call this `name` or `id`? We should probably explain what this is earlier on, it's essentially a label or name for the action.

doc/source/guide/custom.rst
```diff
@@ -0,0 +1,114 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This simple function allows you to remove from Lux's action manager an action with its id. The action will no longer display with the Lux dataframe.
+
+.. code-block:: python
+    
+    lux.remove_action("G10")
+    df
+
+.. image:: https://github.com/lux-org/lux-resources/blob/master/doc_img/custom-4.png?raw=true
```
dorisjlee, line 117: This image seems to belong to earlier instead of here.

dorisjlee (changes requested):

dorisjlee:
Thanks @caitlynachen, this looks great! I made some modifications to `custom.rst` and left some comments. We should incorporate these changes also in the corresponding `8-custom-action.ipynb` file.
I'm adding @jinimukh as a reviewer to do a final check over everything after you are able to address these changes.
@jinimukh: Please read through custom.rst to make sure that the writing makes sense and is easy to understand.

doc/source/guide/custom.rst
```diff
@@ -0,0 +1,123 @@
...
+
+As we can see, Lux displays several recommendation actions, such as Correlation and Distributions, which is globally registered by default.
+
+Registering Custom Actions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let's define a custom function to generate the recommendations on the dataframe. In this example, we register a custom action called `G10` to generate a collection of visualizations that showcases numerical measures that differs significantly across `G10 <https://en.wikipedia.org/wiki/Group_of_Ten_(economics)>`_ and non-G10 countries. In other words, we want to understand how the G10 and non-G10 countries differs based on the measures present in the dataframe. Here, we first generate a list of visualizations (VisList) that looks at how various quantitative attributes breakdown between G10 and non-G10 countries. Then, we score and rank these visualization by calculating the percentage difference in means across G10 v.s. non-G10 countries.
+
+.. code-block:: python
+
```
jinimukh, line 23: Here, you give an example of the type of function that could be a registered action, but can you specify what the return type must be? Is there more formating for this feature, or will all functions that follow this return type work? 

Also, may not be in scope for this PR, but what error handling is there?

doc/source/guide/default.rst
```diff
@@ -0,0 +1,52 @@
...
+In this tutorial, we look at the functionality of the global Config class.
+
+.. code-block:: python
+
+    df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/hpi.csv")
+    df
+
+The dataframe initially registers a few default recommendations, such as Correlation, Enhance, Filter, etc, and defaults to a pandas display.
+
+Registering Custom Actions
```
jinimukh, line 14: This is probably beyond the scope of the PR, but what do you think about making default_display a boolean so that only pandas and lux could be selected (0 and 1 respectively). This removes user error from misspelling "lux" or "pandas" (albeit the possibility of that happening is minuscule). You can maybe make the variable more descriptive -- something like `default_display_to_lux`. Thoughts? @caitlynachen @dorisjlee

jinimukh (approved):

doc/source/guide/default.rst
dorisjlee: That makes sense. This is possibly a good idea.

dorisjlee (approved):

dorisjlee:
Thanks @caitlynachen !

Issue #116, opened 2020-10-18
Better warning message for Vis and VisList


dorisjlee:
We should display a better warning message when the user specifies an intent that indicates more than one item, but puts it into Vis instead of VisList.

For example, if I do: 
```
Vis(["AverageCost","SATAverage", "Geography=?"],df)
```
![image](https://user-images.githubusercontent.com/5554675/96357486-1ba70d80-112f-11eb-8ad9-66f570625fd7.png)
The error that I get is not very interpretable. We should warn users that the intent should be used with a VisList as shown in the screenshot below. If possible, we might even map the intent down to a VisList so that we display something to the user.

![image](https://user-images.githubusercontent.com/5554675/96357500-6b85d480-112f-11eb-8167-bcd40fd07379.png)

cjachekang:
Is an error message like this better? I just raised a syntax error since this behaves like a syntax error.

![image](https://user-images.githubusercontent.com/47467363/98081143-6a40f100-1e2b-11eb-9275-d64915f3c128.png)

dorisjlee:
Yeah, the SyntaxError seems appropriate here. I wonder if we could just map this into a VisList internally and still return something, but place a warning.

cjachekang:
![image](https://user-images.githubusercontent.com/47467363/98150898-37771700-1e84-11eb-95f9-7faeed4ac4db.png)

I've been trying it out and my approach is to construct a VisList if the intent list length is greater than 1 and display that inside of Vis's repr_html - but, it seems to be a bit buggy cause we are trying to have two files import each other. I think we can do it if we turn VisList into something importable (i.e just doing import VisList or something like that) -> I dont know if having two files import each other is good practice though. Alternatively, we can merge Vis and VisList together so that we can construct a VisList inside the Vis and just display the VisList and throw the error after.

dorisjlee:
You're right, it's probably not a good idea to have the two packages import each other. I recall this is one of the reasons why I separated them out in the first place. Let's opt for just showing the error message then, but making it a bit interpretable, something along the lines of: 

> The intent corresponds to more than one visualizations, please replace the `Vis` constructor with `VisList` to generate a list of visualizations. For more information, see: https://lux-api.readthedocs.io/en/latest/source/guide/vis.html#working-with-collections-of-visualization-with-vislist

dorisjlee:
Resolved in #135

dorisjlee:
[...]

PR #135, opened 2020-11-06
Better warning message for Vis and VisList

cjachekang:
Issue: https://github.com/lux-org/lux/issues/116

image


dorisjlee:
Thanks @cjachekang , I made some minor tweaks to the warning message and added a test. However, some of the test cases is not working, and are resulting in other errors, instead of throwing the custom error.

Example 1: 

```
Vis(["SATAverage","?"],df)
```

Example 2:
```
Vis(["SATAverage","AverageCost","Region=New England|Southeast"],df)
```
Could you see if we could get these to show the custom error? We can add  some more examples into `test_multi_vis`.

cjachekang:
Added those two cases to the tests. I moved the SyntaxError check to initialization - probably better to catch the error before all the computations are done.

Also removed traceback for this error so that it looks cleaner - let me know if you want me to add it back.

PR #137, opened 2020-11-11
Fix Inline comments breaking to new lines

jinimukh:
This PR addresses an issue where black reformats arguments to different lines if an inline comment makes line length exceed 79 characters.


lux/processor/Compiler.py
```diff
@@ -395,9 +383,10 @@ def enforce_specified_channel(vis: Vis, auto_channel: Dict[str, str]):
             if len(sAttr) == 1:  # if specified in dobj
                 # remove the specified channel from auto_channel (matching by value, since channel key may not be same)
                 for i in list(auto_channel.keys()):
+                    # need to ensure that the channel is the same (edge case when duplicate Cols with same attribute name)
                     if (auto_channel[i].attribute == sAttr[0].attribute) and (
```
dorisjlee, line 387: Can we rewrite this to be inline?

lux/vis/Vis.py
```diff
@@ -322,9 +322,9 @@ def refresh_source(self, ldf):  # -> Vis:
             from lux.processor.Parser import Parser
```
dorisjlee, line 322: Can we make line 55-59 of Vis.py a single line?

dorisjlee: Simmiliarly, line 80-82 in Vis.py

lux/processor/Compiler.py
jinimukh: I wasn't able to make it inline, but I split it up so that the two conditions are on separate lines

Issue #144, opened 2020-11-17
Speed up test by using shared global variable


dorisjlee:
Use pytest fixture global variable to share dataframes to prevent having to load in the test dataset every time across the different tests. For example: 
![image](https://user-images.githubusercontent.com/5554675/99403865-e9532080-2925-11eb-8df3-c8f67758e01f.png)

This might also help resolve #97 .

dorisjlee:
Resolved in #149, thanks @piyushg9794 !

PR #149, opened 2020-11-19
Global shared variable in test (#144)

piyushg9794:
using pytest fixture global shared variable in test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ static/
 *.egg-info*
 build/
 .DS_Store
+.idea/
 tests/.coverage
 tests/coverage.xml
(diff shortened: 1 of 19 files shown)


dorisjlee:
Thanks for the contribution @piyushg9794!
I made some changes to the PR. In particular, I modified the fixture to share the variable across the session to speed up the test. As a result, I had to resolve several dependencies in cases where the dataframe was being modified (in this case we simply reload the dataframe again), or when the intent is changed (where we simply clear the intent after the test). The runtime of the entire test suite is sped up by around 20 second from this PR.

piyushg9794:
Thanks @dorisjlee for correcting me.

Issue #145, opened 2020-11-17
Improve warning message format


dorisjlee:
The current warning messages via `warnings.warn` is hard to read due to the call trace that is printed alongside the warning. This is true even when we set the `stacklevel` parameter in `warnings.warn`. For example:
![image](https://user-images.githubusercontent.com/5554675/99404699-e60c6480-2926-11eb-8201-acbb6b7755ad.png)
We should add warnings formatted that display nicer warning messages following best practices for warning formatting (refer to warning messages formatting in numpy and pandas?). This could possibly be achieved by a global levels setting or a warnings formatter.

dorisjlee:
Improved warning message stack trace for unexpected error
![image](https://user-images.githubusercontent.com/5554675/100400155-73228c80-3090-11eb-8329-db7adc339476.png)

PR #151, opened 2020-11-20
Reformat Warnings Labels

caitlynachen:
Before: 
image (3)
After:
Screen Shot 2020-11-19 at 3 20 03 PM
Screen Shot 2020-11-19 at 3 33 43 PM


caitlynachen:
https://github.com/lux-org/lux/issues/145

dorisjlee:
Does the `warnings.formatwarning` need to be called in every file or can we make a global call for this inside the `__init__.py` just once so that it is applied across all warnings in the session?

caitlynachen:
I did try to add the import warnings to the init, however since I don't think the imported warnings module can't be passed in throughout different files, the global import didn't work. Let me know if you'd like me to try anything else

dorisjlee:
That makes sense. I was thinking about this in a session's perspective, but I guess there is no sharing of warning across the classes.

dorisjlee:
Could you fix the failing Travis test case?

dorisjlee (approved):

jerrysong1324 (approved):

dorisjlee:
Thanks @caitlynachen !

Issue #129, opened 2020-10-30
Make `plot_config` and choice of renderer a global config 


dorisjlee:
As described [here](https://lux-api.readthedocs.io/en/latest/source/guide/style.html#styling-custom-plot-settings),`plot_config` is used for setting the plotting style for the rendered visualization. Currently, this is a property that is tied with the dataframe, we should extend this to make this a global config setting (e.g., `lux.config.plot_setting`). The configuration is dependent on the choice of renderer too, so it makes sense to have this as a global setting.

dorisjlee:
@jinimukh: Before we close this issue, could we add some documentation for this?

Chandrachud23:
`# Collecting basic usage statistics for Lux`
`lux.logger = True # Remove this line if you do not want your interactions recorded`
`df_train = pd.read_csv('../input/jane-street-market-prediction/train.csv')`
`lux.config.default_display = "lux"`
`df_train`
So while executing this, I found this error that module 'lux' has no attribute 'config'. Can anyone suggest what to do?

dorisjlee:
Hi @Chandrachud23 , `lux.config` is supported only the most recent version of Lux.You can install the latest version by:
```bash
pip install lux-api==0.2.1
jupyter nbextension install --py luxwidget
jupyter nbextension enable --py luxwidget
```
Note that you may have to restart your jupyter notebook after doing this so that the change is reflected.

Note that the lux.config and lux.logger are optional commands. Lux should work as expected if you simply inputted: 
```python
df_train = pd.read_csv('../input/jane-street-market-prediction/train.csv')
df_train
```
Let us know if you are still experiencing problems with this after the suggested fix.

PR #152, opened 2020-11-20
Make plot_config a global variable

jinimukh:
This PR addresses Issue #129.


lux/_config/config.py
```diff
@@ -177,5 +179,50 @@ def default_display(self, type: str) -> None:
                 stacklevel=2,
             )
 
+    @property
+    def default_renderer(self):
```
dorisjlee, line 183: Do we actually need `default_renderer` and `default_plot_config` as additional things inside the config or should we just have this as a variable set from `renderer` and `plot_config`?

lux/_config/config.py
```diff
@@ -177,5 +179,50 @@ def default_display(self, type: str) -> None:
...
+
+        Example
+        ----------
+        Changing the color of marks and adding a title for all charts displayed for this dataframe
+        >>> df = pd.read_csv("lux/data/car.csv")
+        >>> def changeColorAddTitle(chart):
+                        chart = chart.configure_mark(color="red") # change mark color to red
+                        chart.title = "Custom Title" # add title to chart
+                        return chart
+        >>> lux.config.default_plot_config = changeColorAddTitle
```
dorisjlee, line 219: I'm actually a bit confused now of the difference between `lux.config.default_plot_config` and `lux.config.plot_config`.

dorisjlee (changes requested):

lux/_config/config.py
jinimukh: I have changed everything to `lux.config.plot_config`. The internal attribute remains `lux.config._default_plot_config`

PR #157, opened 2020-11-25
Moved Executor Parameters to Global Config

thyneb19:


lux/core/frame.py
```diff
@@ -65,10 +65,10 @@ def __init__(self, *args, **kw):
         self._prev = None
         super(LuxDataFrame, self).__init__(*args, **kw)
 
-        self.executor_type = "Pandas"
-        self.executor = PandasExecutor()
-        self.SQLconnection = ""
-        self.table_name = ""
+        lux.config.executor_type = "Pandas"
+        lux.config.executor = PandasExecutor()
+        lux.config.SQLconnection = ""
```
dorisjlee, line 70: for `frame.py` could we just keep the `lux.config.SQLconnection` and `lux.config.table_name` unspecified and adopt their default values set in the config?

lux/core/frame.py
```diff
@@ -181,12 +181,12 @@ def set_executor_type(self, exe):
                 import psycopg2
```
dorisjlee, line 181: Should we keep `set_executor_type` inside `frame.py` or can we move it all to config or somewhere else?

lux/core/frame.py
```diff
@@ -290,8 +292,8 @@ def __repr__(self):
     #######################################################
 
     def set_SQL_connection(self, connection, t_name):
```
dorisjlee, line 294: Same here, wondering if we should move this outside of `frame.py`?

dorisjlee (changes requested):

dorisjlee:
This looks good, I left some comments on code organization for whether to move things outside of frame.py. We should also update the [config documentation](https://lux-api.readthedocs.io/en/latest/source/reference/config.html) on how to set executor type.

lux/core/frame.py
thyneb19: Originally I set the parameters within config, but importing the PandasExecutor caused some circular dependency so I just set the values within frame.py

thyneb19: SQLconnection's default value is now set in the config. The table_name parameter will remain a frame level parameter to accommodate use case where users work with multiple Lux dataframe each associated with a different database table.

lux/core/frame.py
thyneb19: Since the SQLconnection and executor parameters are now both in config, makes sense to move the set function for them there to.

lux/core/frame.py
jinimukh: maybe you can move it to the global config and put it as a setter?

lux/core/frame.py
```diff
@@ -65,10 +65,8 @@ def __init__(self, *args, **kw):
         self._prev = None
         super(LuxDataFrame, self).__init__(*args, **kw)
 
-        self.executor_type = "Pandas"
-        self.executor = PandasExecutor()
-        self.SQLconnection = ""
         self.table_name = ""
+        lux.config.executor = PandasExecutor()
```
jinimukh, line 69: why do you not need to reset SQLconnection to ""?

jinimukh (approved):
Looks good!

lux/core/frame.py
thyneb19: Yeah I think it would be good to have a setter in the global config for the SQL connection. I think I'd also have to add a separate setter in frame.py for the specific table name so users can associate individual tables with individual data frames still.

dorisjlee:
@thyneb19 : There seem to be some tests that are failing because the executor reference is out of date?

lux/core/frame.py
dorisjlee: Ok let's move these set functions to config then.

dorisjlee (approved):

dorisjlee:
Should be ready for merging once the set functions in `frame.py` are moved to `config.py`

dorisjlee:
Thanks @thyneb19 !

Issue #36, opened 2020-07-17
Incorporate pattern similarity


dorisjlee:
Incorporate pattern similarity into registered action, or show an example of how user-defined action can easily support custom VisList and scoring functions. Adapt existing code in lux/action/similarity.py.

PR #158, opened 2020-11-27
Add similarity as default action

caitlynachen:
Screen Shot 2020-11-27 at 2 49 31 PM


dorisjlee:
Related to #36

lux/action/similarity.py
```diff
@@ -39,11 +39,21 @@ def similar_pattern(ldf, intent, topK=-1):
     recommendations : Dict[str,obj]
         object with a collection of visualizations that result from the Similarity action
     """
-    row_specs = list(filter(lambda x: x.value != "", intent))
-    if len(row_specs) == 1:
-        search_space_vc = VisList(ldf.current_vis.collection.copy(), ldf)
 
-        query_vc = VisList(intent, ldf)
+    last = ldf.intent[-1]
```
dorisjlee, line 43: Why are we picking the last clause in the intent? There is no guarantee that the last element is the one with the value specified. You should fetch the clause by using `utils.get_filter_specs(intent)`.

lux/core/frame.py
```diff
@@ -471,6 +472,7 @@ def maintain_recs(self):
                     lux.register_action("Enhance", enhance, one_current_vis)
                     lux.register_action("Filter", filter, one_current_vis)
                     lux.register_action("Generalize", generalize, one_current_vis)
+                    lux.register_action("Similarity", similar_pattern, one_current_vis, 15)
```
dorisjlee, line 475: Should we change the condition here to be `one_current_vis` and that one of the filter attributes are present, and that we have a line chart being displayed? Otherwise, similarity might not make a lot of sense without a filter attribute or a line chart.

dorisjlee (changes requested):

dorisjlee:
Thanks @caitlynachen, could we add a few tests inside test_action.py that validates that similarity is working correctly? (e.g., checks that the top k items are in fact the most similar ones.)

lux/action/similarity.py
```diff
@@ -39,11 +39,21 @@ def similar_pattern(ldf, intent, topK=-1):
     recommendations : Dict[str,obj]
```
jerrysong1324, line 39: I would change the documentation for the function parameters as well.

lux/action/similarity.py
```diff
@@ -39,11 +39,21 @@ def similar_pattern(ldf, intent, topK=-1):
...
 
-        query_vc = VisList(intent, ldf)
+    last = ldf.intent[-1]
+    query = ldf.intent.copy()[0:-1]
+    query.append(lux.Clause(last.attribute, last.attribute, "?"))
+    row_specs = ldf.intent
+    if len(args) == 0:
+        topK = 15
+    else:
+        topK = args[0][0]
```
jerrysong1324, line 50: I'm not too sure how the `args` will be formatted, but I'm a little concerned that while `arg[0]` will not be null due to your previous check on length, there might not be a value for `args[0][0]`

lux/action/similarity.py
```diff
@@ -55,7 +65,7 @@ def similar_pattern(ldf, intent, topK=-1):
             preprocess(vis)
             vis.score = euclidean_dist(query_vis, vis)
         search_space_vc.normalize_score(invert_order=True)
-        if topK != -1:
+        if(topK != -1):
```
jerrysong1324, line 68: Nit: You don't really need parenthesis here : )

jerrysong1324 (commented):
Are there tests on this?

dorisjlee:
@caitlynachen: Could you fix the failing test on this before we review the PR?
Also, could you please remove the test notebook `examples/similarity.ipynb` and move the examples into the test files?
Thanks!

PR #160, opened 2020-11-28
Update Makefile and pyproject.toml

jinimukh:
1. Add `pyproject.toml` which can hold all configurations for `black`. Thus, we can run just `black .` as we did before and get the correct line length.

2. Update the Makefile so running `make test` will run the same commands as Travis.


dorisjlee:
Thanks @jinimukh !

PR #174, opened 2020-12-04
Fix bug caused by `groupby.agg` on column with many unique values

westernguy2:
For the example in `cars.csv`, where we run `df.groupby("Brand").agg(sum)`, we were getting a `KeyError`. This is now fixed. This was caused by a combination of factors. First, the `Clause` objects created in `column_group.py` were setting `aggregation` to None instead of "". Second, `execute_aggregation` was dropping the `index` column in vis data, even though it was a necessary part of the vis data. This is also now patched as well. 

I've added a test for this specific case as well.


dorisjlee:
I'm not sure if changing aggregation from None to "" is the right approach.
The aggregation as "None" is a special keyword to ensure that the aggregation is not performed (see [here](https://github.com/lux-org/lux/blob/da4c6027c5467a10b646ddb0bb50f3549fd38acc/lux/executor/PandasExecutor.py#L134) and the [docstring for Clause](https://github.com/lux-org/lux/blob/master/lux/vis/Clause.py#L62)), since generally aggregation as "" means that the aggregation is set as the default value (i.e., mean). It might be better to stick with one or the other.

dorisjlee:
I added the fix also for row group and added a test. This is the Year visualization before the fix:
image
this is after the fix:
image
In general, we should keep the row and column group as symmetrical as possible (transpose is a good way of testing this).

dorisjlee (approved):

dorisjlee (approved):

PR #182, opened 2020-12-19
Similarity as a default action

caitlynachen:
Screen Shot 2020-12-19 at 2 10 17 PM
https://github.com/lux-org/lux/pull/158 - original PR set on separate branch, had to update to master

Original issue #36

### DIFF
--- a/lux/action/similarity.py
+++ b/lux/action/similarity.py
@@ -19,6 +19,7 @@ import numpy as np
 from lux.vis.VisList import VisList
 from lux.utils.utils import get_filter_specs
 
+
 def aggregate(vis):
     """
     Aggregates data values on the y axis so that the vis is a time series
--- a/lux/action/similarity.py
+++ b/lux/action/similarity.py
@@ -61,13 +61,14 @@ def similar_pattern(ldf, *args):
             "action": "Similarity",
             "description": "Show other charts that are visually similar to the Current vis.",
         }
-        for vis in search_space_vc:
-            preprocess(vis)
-            vis.score = euclidean_dist(query_vis, vis)
+        search_space_vc_copy = VisList(query, ldf)
+        for i in range(len(search_space_vc_copy)):
+            preprocess(search_space_vc_copy[i])
+            search_space_vc[i].score = euclidean_dist(query_vis, search_space_vc_copy[i])
         search_space_vc.normalize_score(invert_order=True)
         if topK != -1:
             search_space_vc = search_space_vc.topK(topK)
-        recommendation["collection"] = search_space_vc
+        recommendation["collection"] = search_space_vc[1:]
         return recommendation
     else:
         print("Query needs to have 1 row value")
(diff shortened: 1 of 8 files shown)


tests/test_action.py
```diff
@@ -197,3 +197,23 @@ def test_year_filter_value(global_var):
...
+            lux.Clause("Year", channel="x"),
+            lux.Clause("Displacement", channel="y"),
+            lux.Clause("Origin=USA"),
+        ]
+    )
+    df._repr_html_()
+    assert len(df.recommendation["Similarity"]) == 3
+    assert (
+        df.recommendation["Similarity"][0].score
+        > df.recommendation["Similarity"][1].score
```
dorisjlee, line 216: This is not really testing anything regarding the similarity metric, since if we are sorting correctly, the first score should always be bigger than the second, than the third. What we should have is having the visualization that we know "look" similar and compare the score of those instead.

lux/action/similarity.py
```diff
@@ -17,9 +17,10 @@
 import math
 import numpy as np
 from lux.vis.VisList import VisList
+from lux.utils.utils import get_filter_specs
 
 
-def similar_pattern(ldf, intent, topK=-1):
+def similar_pattern(ldf, *args):
```
dorisjlee, line 23: Merge with `Filter` action and move the similarity metric into an interestingness metric

dorisjlee:
Thanks @caitlynachen. I went through the code and added some tests. Here are some issues that we should address: 
1) There is a bug where the `normalize` causes the y-axes values to be between 0 and 1. The interestingness score calculation should not modify the `vis.data`, but simply be used to compute the scores used for ranking.
Screen Shot 2020-12-23 at 4 49 43 PM
2) We should remove the top-ranked item, which is generally the same thing as what is specified in the Current Visualization. (e.g., City=Crofton in this case)
image
3) Related to (1), we should make the similarity more of an interestingness metric and apply this interestingness metric and merge this with the `Filter` action when the similarity condition is met.

tests/test_interestingness.py
```diff
@@ -238,12 +237,7 @@ def test_interestingness_0_2_0(global_var):
             and str(df.recommendation["Enhance"][f].mark) == "scatter"
         ):
             rank2 = f
-        if (
-            str(df.recommendation["Enhance"][f]._inferred_intent[2].attribute) == "Year"
-            and str(df.recommendation["Enhance"][f].mark) == "scatter"
-        ):
-            rank3 = f
-    assert rank1 < rank2 and rank1 < rank3 and rank2 < rank3
+    assert rank1 < rank2
```
thyneb19, line 240: Is there a reason why the rank3 check was removed here? Was wondering if that specific chart was no longer being included in the Enhance action's recommendation list.

dorisjlee: I removed this because we are invalidating charts that have 2 measures and 1 temporal value (it doesn't make sense to "color" a temporal value), so this chart is no longer showing up in the Enhance recommendation list.

caitlynachen: this was reverted:)

dorisjlee:
I made some minor cleanup and changed the similarity metric from -1 * distance to 1- distance to avoid negative scores. Great work @caitlynachen!

PR #185, opened 2020-12-23
Coalesce all data_type attributes of frame into one

jinimukh:
This PR standardizes data_type attributes into data_type_lookup and implements methods to compute data_type_lookup, data_model, and data_model_lookup

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -21,6 +21,7 @@ from lux.history.history import History
 from lux.utils.message import Message
 from lux.utils.utils import check_import_lux_widget
 from typing import Dict, Union, List, Callable
+
 # from lux.executor.Executor import *
 import warnings
 import traceback
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -21,6 +21,7 @@ from lux.history.history import History
 from lux.utils.message import Message
 from lux.utils.utils import check_import_lux_widget
 from typing import Dict, Union, List, Callable
+from lux.executor.Executor import *
 import warnings
 import traceback
 import lux
@@ -36,9 +37,6 @@ class LuxDataFrame(pd.DataFrame):
         "_intent",
         "_inferred_intent",
         "data_type_lookup",
-        "data_type",
-        "data_model_lookup",
-        "data_model",
         "unique_values",
         "cardinality",
         "_rec_info",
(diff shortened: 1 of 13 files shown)


lux/core/frame.py
```diff
@@ -127,14 +122,25 @@ def expire_metadata(self):
...
         self.data_type_lookup = None
-        self.data_type = None
-        self.data_model_lookup = None
-        self.data_model = None
         self.unique_values = None
         self.cardinality = None
         self._min_max = None
         self.pre_aggregated = None
 
+    def compute_data_type_from_lookup(self):
```
dorisjlee, line 130: Why is this inside the `frame.py` instead of the `Executor.py`(which is still shared across `SQLExecutor` and `PandasExecutor`)? I think that earlier this semester @thyneb19 worked on moving data type related stuff outside of `frame.py` so maybe we should keep it that way?

tests/test_type.py
```diff
@@ -37,7 +37,9 @@ def test_check_int_id():
         "https://github.com/lux-org/lux-datasets/blob/master/data/instacart_sample.csv?raw=true"
     )
     df._repr_html_()
-    assert len(df.data_type["id"]) == 3
+    data_type = df.compute_data_type_from_lookup()
+    assert len(data_type["id"]) == 3
+    # assert len(df.data_type["id"]) == 3
```
dorisjlee, line 42: remove comment?

tests/test_pandas_coverage.py
```diff
@@ -75,19 +84,28 @@ def test_rename(global_var):
...
+
+    data_model_lookup = df.compute_data_model_lookup()
+    new_data_model_lookup = new_df.compute_data_model_lookup()
 
-    assert df.data_type["nominal"][0] == "Name"
-    assert new_df.data_type["nominal"][0] == "Car Name"
+    assert data_model_lookup != new_data_model_lookup
 
-    assert df.data_model_lookup != new_df.data_model_lookup
+    assert data_model_lookup["Name"] == new_data_model_lookup["Car Name"]
```
dorisjlee, line 100: Here maybe we could just compute the equality between the `data_model`s

dorisjlee (changes requested):

lux/core/frame.py
jinimukh: Great, I'll move it over

tests/test_pandas_coverage.py
jinimukh: So, do you want me to get rid of lines 98-100?

dorisjlee (approved):

dorisjlee (approved):

Issue #181, opened 2020-12-19
[BUG] User-provided Index causes KeyError in Pandas Execution


briankim1027:
1. i installed as follow guideline in jupyter notebook cell
    !pip install lux-api
    !jupyter nbextension install --py luxwidget
    !jupyter nbextension enable --py luxwidget

2. import lux and read csv file then call dataframe
   i got an message as follows.  

how can i fix it? thank U

C:\Users\brian\anaconda3\lib\site-packages\IPython\core\formatters.py:345: UserWarning:
Unexpected error in rendering Lux widget and recommendations. Falling back to Pandas display.
Please report the following issue on Github: https://github.com/lux-org/lux/issues 

C:\Users\brian\anaconda3\lib\site-packages\lux\core\frame.py:711: UserWarning:Traceback (most recent call last):
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\core\frame.py", line 663, in _repr_html_
    self.maintain_recs()
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\core\frame.py", line 494, in maintain_recs
    custom_action_collection = custom_actions(rec_df)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\action\custom.py", line 73, in custom_actions
    recommendation = lux.actions.__getattr__(action_name).action(ldf, args)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\action\univariate.py", line 84, in univariate
    vlist = VisList(intent, ldf)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\vis\VisList.py", line 43, in __init__
    self.refresh_source(self._source)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\vis\VisList.py", line 313, in refresh_source
    ldf.executor.execute(self._collection, ldf)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\executor\PandasExecutor.py", line 97, in execute
    PandasExecutor.execute_aggregate(vis, isFiltered=filter_executed)
  File "C:\Users\brian\anaconda3\lib\site-packages\lux\executor\PandasExecutor.py", line 171, in execute_aggregate
    vis._vis_data = vis.data[[groupby_attr.attribute, "Record"]]
  File "C:\Users\brian\anaconda3\lib\site-packages\pandas\core\frame.py", line 2912, in __getitem__
    indexer = self.loc._get_listlike_indexer(key, axis=1, raise_missing=True)[1]
[...]

PR #191, opened 2020-12-27
Bug Fix: User-provided Index causes KeyError in Pandas Execution

thyneb19:
Addressing #181. Fix bug in Pandas executor caused when user specifies index column. Added test cases for `set_index`.


lux/executor/PandasExecutor.py
```diff
@@ -156,20 +156,28 @@ def execute_aggregate(vis: Vis, isFiltered=True):
...
                 if has_color:
                     vis._vis_data = (
                         vis.data.groupby([groupby_attr.attribute, color_attr.attribute], dropna=False)
                         .count()
                         .reset_index()
                     )
-                    vis._vis_data = vis.data.rename(columns={"index": "Record"})
+                    index_name = list(
+                        filter(
+                            lambda k: groupby_attr.attribute not in k and color_attr.attribute not in k,
```
westernguy2, line 167: Why do we do `groupby_attr.attribute not in k` does that check if the groupby attribute is in the column name `k`? Why not just directly check if they're equal?

lux/executor/PandasExecutor.py
```diff
@@ -156,20 +156,28 @@ def execute_aggregate(vis: Vis, isFiltered=True):
...
+                    )[0]
+                    vis._vis_data = vis.data.rename(columns={index_name: "Record"})
                     vis._vis_data = vis.data[[groupby_attr.attribute, color_attr.attribute, "Record"]]
                 else:
                     vis._vis_data = (
                         vis.data.groupby(groupby_attr.attribute, dropna=False).count().reset_index()
                     )
-                    vis._vis_data = vis.data.rename(columns={"index": "Record"})
+                    index_name = list(
+                        filter(lambda k: groupby_attr.attribute not in k, vis.data.columns)
```
westernguy2, line 178: Same question as above here

lux/executor/PandasExecutor.py
thyneb19: Yeah here I'm trying to get the name of the column with the aggregate counts, which is named either "index" or whatever column a user has set to be the index. Ideally I would like to get the index name from the original Lux DF using something like `ldf.index.name` but the PandasExecutor doesn't have a reference to this DF, so we'd have to pass down the index name some other way. Instead of adding another parameter to the global config just for this, I just filtered out the groupby and color attributes to get at the index column.

dorisjlee: I think this makes sense but is a bit messy. We should look into whether there is a way of rewriting the previous line (possibly by replacing `.count` via `.transform` or `.agg` instead, or something via the reset_index) so that we can explicitly do the renaming to `Record` here. Something like [this](https://stackoverflow.com/questions/40914200/can-i-assign-a-reset-index-a-name) might work.
```python
vis._vis_data = (
                        vis.data.groupby([groupby_attr.attribute, color_attr.attribute], dropna=False)
                        .count()
                        .reset_index()
                    )
```

lux/executor/PandasExecutor.py
dorisjlee: same suggestion as above

dorisjlee (changes requested):

lux/executor/PandasExecutor.py
dorisjlee: My concern for the `index_name = list( filter(lambda k: groupby_attr.attribute not in k, vis.data.columns)[0])` is that I'm not sure if there would ever be an edge case where there is no matching filter and this causes an IndexError. So if we can combine it with the previous statement, it would be better.

dorisjlee:
This looks good, I've run some tests and it is largely working. Attached some suggestions on improving the code.

lux/executor/PandasExecutor.py
thyneb19: Thanks for the rename suggestion, that solution seems to work out. I also found that I vis.data still has the index name so I can use that instead of filtering out the non-index columns.

westernguy2 (approved):

dorisjlee (approved):

Issue #183, opened 2020-12-21
[ENH] Make sampling parameter a global config


bilalmussaukfd:
Hiya
Is it possible to remove the 30000 rows random sampling so that I can see the entirety of my data albeit it will be a bit slow to process?

![image](https://user-images.githubusercontent.com/75366970/102788267-4402f080-439a-11eb-8db3-a5997c2f3a73.png)

dorisjlee:
Hi @bilalmussaukfd, 

Yeah, we can definitely add this in as a configuration parameter that you can turn on and off. 
We can also expose a parameter to let you adjust the number of samples and the number of rows that we chose to sample.
As you mentioned, the main concern is that turning it off might make things run really slow, especially when lots of things are plotted on the frontend. We can work on this and include this as an optional configuration in the next release. 

As a temporary patch, if you want to try turning off the sampling right now, you can simply replace [this one line inside PandasExecutor.py](https://github.com/lux-org/lux/blob/master/lux/executor/PandasExecutor.py#L83) with `ldf._sampled = ldf`, as follows: 
```python
#PandasExecutor.execute_sampling(ldf)
ldf._sampled = ldf
```
If you are not sure where your lux is installed, you can print out the source location by:
```python
import lux
lux.__file__
```
and then access `lux/executor/PandasExecutor.py`. The changes should be reflected after you save the source and restart your notebook to import lux again. Let us know if you have any questions in the meanwhile. Thanks!

bilalmussaukfd:
Hi Doris,
I have made that change but it has slowed it down drastically. My dataset is only 34k (I trimmed it down from the initial 587k rows) rows so not sure if I am testing its limits?
Is Lux better for smaller datasets in this instance?

Regards

Bilal

dorisjlee:
Hi @bilalmussaukfd, 

[...]

PR #192, opened 2020-12-29
Add sampling parameters as a global config

westernguy2:
Allows users to enable/disable sampling by running, for example, `lux.config.sampling = False`. Users can also edit the `sampling_start` and `sampling_cap` parameters in a similar fashion for sampling large dataframes. 

Also, added tests for this new feature, and edited the documentation/FAQ. 

This solves the issue in #183


dorisjlee (approved):

dorisjlee:
This looks great, I've made minor changes to the docs, but otherwise, the PR looks good. While we're at this, could we also add a disable/enable option for whether a heatmap/scatterplot is generated (and associated test + docs)? I think we just need to modify it [here](https://github.com/lux-org/lux/blob/master/lux/executor/PandasExecutor.py#L101)

dorisjlee:

PR #198, opened 2021-01-06
Matplotlib rendering option

caitlynachen:
Screen Shot 2021-01-06 at 10 41 11 AM
Screen Shot 2021-01-06 at 10 40 08 AM
Screen Shot 2021-01-06 at 10 40 00 AM
Screen Shot 2021-01-06 at 10 39 54 AM
Screen Shot 2021-01-06 at 10 39 50 AM


dorisjlee (commented):
Can we remove the `.coverage` file?

lux/_config/config.py
```diff
@@ -26,6 +26,7 @@ def __init__(self):
         self._sampling_cap = 30000
         self._sampling_flag = True
         self._heatmap_flag = True
+        self._set_vislib = "vegalite"
```
dorisjlee, line 29: Could we rename this to `plotting_backend` so that it uses the same terminology as [pandas's `pd.options.plotting.backend`](https://plotly.com/python/pandas-backend/)?
So users can change the visualization library used via `lux.config.plotting_backend='matplotlib'`.

lux/_config/config.py
```diff
@@ -133,6 +134,29 @@ def default_display(self, type: str) -> None:
...
+        type : str
+                Default display type, can take either the string `lux` or `pandas` (regardless of capitalization)
+        """
+        if type.lower() == "vegalite":
+            self._set_vislib = "vegalite"
+        elif type.lower() == "matplotlib":
+            self._set_vislib = "matplotlib"
+        else:
+            warnings.warn(
+                "Unsupported vislib type. Vislib should either be `vegalite` or `matplotlib`.",
```
dorisjlee, line 156: `Unsupported plotting backend. Lux currently only support 'altair', 'vegalite', or 'matplotlib'`.

lux/_config/config.py
```diff
@@ -133,6 +134,29 @@ def default_display(self, type: str) -> None:
...
+    @set_vislib.setter
+    def set_vislib(self, type: str) -> None:
+        """
+        Set the widget display to show Pandas by default or Lux by default
+        Parameters
+        ----------
+        type : str
+                Default display type, can take either the string `lux` or `pandas` (regardless of capitalization)
+        """
+        if type.lower() == "vegalite":
```
dorisjlee, line 150: maybe both `'altair'` and `'vegalite'` should point to the same thing?

lux/vis/Vis.py
```diff
@@ -236,6 +236,46 @@ def to_Altair(self, standalone=False) -> str:
         self._code = renderer.create_vis(self, standalone)
         return self._code
 
+    def to_Matplotlib(self, standalone=False) -> str:
```
dorisjlee, line 239: switch to lowercase M for matplotlib

dorisjlee, line 239: Where is the `standalone` parameter being use? If not, then remove from source.

lux/vis/Vis.py
```diff
@@ -236,6 +236,46 @@ def to_Altair(self, standalone=False) -> str:
...
+        """
+        Generate minimal Altair code to visualize the Vis
```
dorisjlee, line 241: Update docs here and throughout. Still mentions Altair.

lux/vis/Vis.py
```diff
@@ -236,6 +236,46 @@ def to_Altair(self, standalone=False) -> str:
...
+        str
+                String version of the Altair code. Need to print out the string to apply formatting.
+        """
+        from lux.vislib.matplotlib.MatplotlibRenderer import MatplotlibRenderer
+
+        renderer = MatplotlibRenderer(output_type="matplotlib")
+        self._code = renderer.create_vis(self, standalone)
+        return self._code
+
+    def to_Matplotlib_code(self, standalone=False) -> str:
```
dorisjlee, line 259: lowercase M

lux/vislib/matplotlib/BarChart.py
```diff
@@ -0,0 +1,120 @@
...
+        plt.tight_layout()
+
+        # Convert chart to HTML
+        import base64
+        from io import BytesIO
+
+        tmpfile = BytesIO()
+        self.fig.savefig(tmpfile, format="png")
+        chart_code = base64.b64encode(tmpfile.getvalue()).decode("utf-8")
+        # Inside chartGallery.tsx change VegaLite component to be adaptable to different rendering mechanism (e.g, img)
```
dorisjlee, line 104: remove comment

lux/vislib/matplotlib/BarChart.py
```diff
@@ -0,0 +1,120 @@
...
+        objects = df[bar_attr].astype(str)
+        performance = df[measure_attr]
+
+        self.ax.barh(objects, performance, align="center")
+        self.ax.set_xlabel(x_attr_abv)
+        self.ax.set_ylabel(y_attr_abv)
+        plt.tight_layout()
+
+        # Convert chart to HTML
+        import base64
```
dorisjlee, line 98: Line 98-103 repeated across all charts. Extract to shared function inside MatplotlibChart or MatplotlibRenderer to enable reuse?

lux/vislib/matplotlib/Heatmap.py
```diff
@@ -0,0 +1,93 @@
...
+        from io import BytesIO
+
+        tmpfile = BytesIO()
+        self.fig.savefig(tmpfile, format="png")
+        chart_code = base64.b64encode(tmpfile.getvalue()).decode("utf-8")
+        # Inside chartGallery.tsx change VegaLite component to be adaptable to different rendering mechanism (e.g, img)
+        # '<img src=\'data:image/png;base64,{}\'>
+
+        self.code += "import matplotlib.pyplot as plt\n"
+        self.code += "import seaborn as sns\n"
```
dorisjlee, line 74: possible to remove seaborn as dependency and use plt.imshow?

dorisjlee (commented):
We may want to incorporate some of the changes inside the plotting libraries that I did for #203, in order to support columns with integer values.

lux/vislib/matplotlib/Histogram.py
```diff
@@ -0,0 +1,98 @@
...
+            msr_attr_abv = msr_attr.attribute[:10] + "..." + msr_attr.attribute[-7:]
+
+        x_min = self.vis.min_max[msr_attr.attribute][0]
+        x_max = self.vis.min_max[msr_attr.attribute][1]
+
+        # msr_attr.attribute = msr_attr.attribute.replace(".", "")
+
+        x_range = abs(max(self.vis.data[msr_attr.attribute]) - min(self.vis.data[msr_attr.attribute]))
+        plot_range = abs(x_max - x_min)
+        markbar = x_range / plot_range * 12
```
dorisjlee, line 54: For histograms, we can decrease the bin width, the bars look a bit thick. It seems like `markbar` isn't being used here. maybe matplotlib has a better mechanism for determining histogram bar width, number of bins, and spacing?

tests/test_matplotlib.py
```diff
@@ -0,0 +1,260 @@
+#  Copyright 2019-2020 The Lux Authors.
```
dorisjlee, line 1: Why were these specific tests chosen? Can we add to `test_vis.py` and have one test for each chart type? Then run them through both Altair and Matplotlib for checks?

dorisjlee (changes requested):

dorisjlee (commented):
For the plotting aesthetics, it would help to have larger fonts on the axes labels and titles. Finally, can we add documentation to FAQ and Config on this?

dorisjlee:
Thanks @caitlynachen, this looks great! I left some comments to address, but I think we are pretty close!

lux/vislib/matplotlib/Histogram.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.code += "import matplotlib.pyplot as plt\n"
+        self.code += "import numpy as np\n"
+
+        self.code += f"df = pd.DataFrame({str(self.data.to_dict())})\n"
+
+        self.code += f"fig, ax = plt.subplots()\n"
+        self.code += f"objects = df['{msr_attr.attribute}']\n"
+
+        self.code += f"counts, bins = np.histogram({str(self.data.to_dict())})\n"
```
dorisjlee, line 94: synthesized code for histogram is not working.

dorisjlee (commented):
> Resolved

lux/vislib/matplotlib/ScatterChart.py
```diff
@@ -0,0 +1,87 @@
...
+        objects = df[x_attr.attribute]
+        performance = df[y_attr.attribute]
+
+        self.ax.scatter(objects, performance)
+        self.ax.set_xlabel(x_attr_abv)
+        self.ax.set_ylabel(y_attr_abv)
+        plt.tight_layout()
+
+        # Convert chart to HTML
+        import base64
```
jerrysong1324: Nit: move this to the top

tests/test_matplotlib.py
```diff
@@ -0,0 +1,260 @@
...
+        [
+            lux.Clause(attribute="Horsepower"),
+            lux.Clause(attribute="MilesPerGal"),
+            lux.Clause(attribute="Acceleration", filter_op=">", value=10),
+        ]
+    )
+    lux.config.set_vislib = "matplotlib"
+    df._repr_html_()
+
+    from lux.utils.utils import get_filter_specs
```
jerrysong1324: Nit: move to the top

jerrysong1324 (commented):
Great work! Adding support for Matplotlib will address a great number of user concerns/feedback. I loosely reviewed the code and the major improvement I would add is some error or feedback to user if they input some graphing package that we do not support yet. Everything else is stye nitpick points.

jerrysong1324:
I will work on testing via the front end to see if I can find bugs / see other issues now.

dorisjlee:
Thanks for all the improvements @caitlynachen! I noted down a couple aesthetics concerns, but the programming logic largely looks good. Most of these concerns are largely related to how the fixed figsize "squishes" the plot, we might be able to find a way of making the figure size a bit more adaptive (longer horizontally is okay in certain scenarios like in the case of a legend). We should also make the plot style (font, size, weight) as similar to what we have in Altair as possible.

1) When the values are long, the plot gets squashed. It looks like Vega-Lite automatically applies some abbreviation, which matplotlib doesn't do. So we should try to accommodate for this.
image
2) Can we make the x and y titles as well as the filter plot titles larger for all charts and with thicker fonts (to distinguish it from the labels)? This can be shared across all charts inside `apply_default_config` inside Matplotlib. We can apply a similar default config as the plot style for AltairChart (see [here](https://github.com/lux-org/lux/blob/master/lux/vislib/altair/AltairChart.py#L52)).
image
3) Increase the font weight/thickness for the orange text ("+2 more...") so that it is more visible. Might also need to increase the font size and move it up a bit too.
image
4) One really cool thing is that we can actually apply seaborn externally, and have it apply to all of our plots! The histograms look a bit squished here, but otherwise, it is pretty nice.
image
5) For colored charts, let's use a [qualitative color map](https://matplotlib.org/3.1.0/tutorials/colors/colormaps.html) (e.g., Paired, Set1) for the categorical color variables so that the categories are more distinguished.
image
6) For the color legend, we should increase the allocated figure size, so that the plot doesn't get squished to the side because of the legend. Similar to Altair: 
image

lux/core/frame.py
```diff
@@ -745,7 +745,7 @@ def current_vis_to_JSON(vlist, input_current_vis=""):
         current_vis_spec = {}
         numVC = len(vlist)  # number of visualizations in the vis list
         if numVC == 1:
```
jerrysong1324, line 747: Is it possible here for the vlist to be of length 0?

dorisjlee: I don't think so, since we only call this when self.current_vis is not None (see line [729](https://github.com/lux-org/lux/pull/198/files#diff-b3f5355f16af7450f6bf0669466ac57ae020ed8c0a56105e8ece9c8c035a8879R729))

caitlynachen: Yup this was how it was done in the past as well

dorisjlee:
Thanks for all the changes @caitlynachen, @jerrysong1324! 
The new changes looks really good, I've tested the code changes and the only issue that I found was some problems with displaying long text columns:
```python 
df = pd.read_csv("https://github.com/chiphuyen/just-pandas-things/blob/master/data/interviews.csv?raw=True")
df['Date'] = pd.to_datetime(df['Date'], format='%b %d, %Y')
```
image
image

The scrolling estimate is a bit off, but we can address this separately in [lux-widget issue #65](https://github.com/lux-org/lux-widget/issues/65).
image

PR #211, opened 2021-01-09
Version lock for jupyter-client

dorisjlee:
An update for jupyter-client breaks the build process on Travis (and other Linux systems), due to this new change that [enforces a minimum requirement on jedi](https://github.com/jupyter/jupyter_client/pull/596). We add a version lock for jupyter-client (v6.1.6) to avoid this issue. This is also implemented in [lux-widget](https://github.com/lux-org/lux-widget), where the error is showing up.
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - "3.7"
 install:
-  - pip install -r requirements.txt
   - pip install -r requirements-dev.txt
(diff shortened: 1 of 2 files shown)


davidbrochart:
FYI, jupyter-client v6.1.11 is released, which removes the jedi pinning. Sorry for the inconvenience.

dorisjlee:
Thanks for letting us know @davidbrochart!

Issue #99, opened 2020-09-27
API support for type override


dorisjlee:
Occasionally, there might be mistakes in the data type that is automatically detected in Lux, we should support the ability to let users override the inferred data type and persist this inferred setting throughout the session (even if the metadata has to be recomputed again).

dorisjlee:
For example, in the real estate data, the `Year` and `Month` attributes are detected as temporal, but they are not true datetimes, but a sequence of ordered months and years. User-defined type overriding would be useful here.
```
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/real_estate_tutorial.csv?raw=true")
# df["Year"] = pd.to_datetime(df["Year"], format='%Y') 
# df["Month"] = pd.to_datetime(df["Month"], format='%M')
```

dorisjlee:
Pending documentation and tutorial on data types before we close the issue.

PR #215, opened 2021-01-11
Explicit data type override via df.set_data_type

jinimukh:
This PR addresses #99 and implements the option for the user to specify lux type in case our inferred type is incorrect.

Changed:
- implementation for df.set_data_type: includes a call to df.maintain_metadata() if relevant metadata does not yet exist.
- tests for correctness and failure cases
- update warning thrown when Lux detects temporal warnings to also include set_data_type suggestion.


lux/core/frame.py
```diff
@@ -242,6 +244,26 @@ def set_intent_as_vis(self, vis: Vis):
         self._intent = vis._inferred_intent
         self._parse_validate_compile_intent()
 
+    def asLuxType(self, types: dict):
```
thyneb19, line 247: Do we have a way to check if users have specified valid datatypes? Wondering if there should be some quick validation to see if all the types set by a user are either nominal, quantitative, id, or temporal.

jinimukh: Ahhh, great catch. I am now throwing an error if it is not in those 4. How does that sound?

thyneb19: Yup the new changes look good!

thyneb19 (approved):

dorisjlee (approved):

dorisjlee:
Thanks @jinimukh, this looks great! I've tested the code and made minor changes to the docs and warnings. 
Once we add some documentation around this feature, we should be able to merge in the changes.

dorisjlee:
Thanks @jinimukh !

PR #231, opened 2021-01-15
[PATCH] Make df.data_type a property that accesses df._data_type

jinimukh:
Before, the user could change `df.data_type` by setting it to some other dictionary. This would benefit from being a "protected variable". Thus, `df.data_type` is a property getter for `df._data_type`. 

Also, `df.data_type` only returned a non-null dictionary if `df._repr_html()` was called before. Now, the getter calls `df.maintain_metadata()` if `df._data_type` is null.

Changed:
- changed metadata attribute `df.data_type` to `df._data_type`
- added a getter that calls `df.maintain_metadata` if `df._data_type` is NULL

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -313,7 +313,7 @@ class LuxDataFrame(pd.DataFrame):
         self.get_SQL_unique_values()
         # self.get_SQL_cardinality()
         for attribute in self.columns:
-            if self.data_type[attribute] == "quantitative":
+            if self._data_type[attribute] == "quantitative":
                 self._min_max[attribute] = (
                     self[attribute].min(),
                     self[attribute].max(),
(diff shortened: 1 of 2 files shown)


lux/core/frame.py
```diff
@@ -101,6 +101,12 @@ def f(*args, **kwargs):
     def history(self):
         return self._history
 
+    @property
+    def data_type(self):
+        if not self._data_type:
+            self.maintain_metadata
```
dorisjlee, line 107: This line needs to be actually called right?
`self.maintain_metadata()`

dorisjlee (changes requested):

dorisjlee:
This looks good @jinimukh, could you resolve the comment and branch conflict before we merge? Thanks!

lux/core/frame.py
jinimukh: Oooooops -- fixed in the next push

dorisjlee (approved):

PR #232, opened 2021-01-15
Temporal type detection for integers via PandasExecutor._is_datetime_number

### LINKED ISSUE #218: Basic Regex-based temporal type detection
Currently, we detect something as temporal data type if its column name is one of these ` ["month", "year", "day", "date", "time"]`, if it is a `pd.Timestamp` object or if it is a string that is convertible to a datetime. This means that if a number seems like a year, month or day, it would not be converted into datetime. (Source [here](https://github.com/lux-org/lux/blob/master/lux/executor/PandasExecutor.py#L402))

For example, if we change the `Year` column in the cars dataset to `blah`, it will no longer be detected as "temporal"
```python
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/car.csv?raw=True")
df = df.rename(columns={"Year":"blah"})
df
```
We could implement an extra layer of basic Regex to check for these cases when the numbers look very similar to a datetime field.
image

Some test data can be found [here](https://github.com/lux-org/lux-datasets), feel free to find your own on Kaggle or elsewhere. For a pull request, please include [tests](https://github.com/lux-org/lux/blob/master/tests/test_type.py) to try out the bugfix on several datasets to verify that temporal fields are being detected and that non-temporal fields are not detected.

micahtyong:
## In this PR

Closes #218 by adding a datetime validator, `_is_datetime_number(series)`, which takes in a series, checks if it's numerical, and outputs whether the numerical values (e.g. year, date) can be interpreted as datetimes. 
- It does so without the need for Regex, by using the Pandas built-in `to_datetime()` function which raises an exception if the input is likely, not interpretable as a datetime (e.g. year is way past the current date, the number is negative). 
- Numerical values representing the day of the week or month of the year are currently not interpreted as a temporal type unless the column name is "day" or "month". The argument here is that it's difficult to conclude that such a common range of values (e.g 1 - 30) are absolutely temporal.  

## Changes
- Implementation for `PandasExecutor._is_datetime_number()`. 
- Tests in `tests/test_type.py` for correctness and failure cases

## Example Output
Screen Shot 2021-01-15 at 1 19 21 PM

### DIFF
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -389,13 +389,11 @@ class PandasExecutor(Executor):
         ldf.data_type = {}
         self.compute_data_type(ldf)
 
-    
-
     def compute_data_type(self, ldf: LuxDataFrame):
         from pandas.api.types import is_datetime64_any_dtype as is_datetime
 
         for attr in list(ldf.columns):
-            temporal_var_list = ["month", "year", "day", "date", "time"]
+            temporal_var_list = ["month", "year", "day", "date", "time", "weekday"]
             if is_datetime(ldf[attr]):
                 ldf.data_type[attr] = "temporal"
             elif self._is_datetime_string(ldf[attr]):
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -211,7 +211,10 @@ class PandasExecutor(Executor):
                             }
                         )
                         vis._vis_data = vis.data.merge(
-                            df, on=[columns[0], columns[1]], how="right", suffixes=["", "_right"],
+                            df,
+                            on=[columns[0], columns[1]],
+                            how="right",
+                            suffixes=["", "_right"],
                         )
                         for col in columns[2:]:
                             vis.data[col] = vis.data[col].fillna(0)  # Triggers __setitem__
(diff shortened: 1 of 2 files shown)


micahtyong:
cc: @dorisjlee. Just worked through [this issue](https://github.com/lux-org/lux/issues/218)! Lux can now classify numerical columns as temporal. I hope this helps!

lux/executor/PandasExecutor.py
```diff
@@ -473,10 +468,22 @@ def _is_datetime_string(series):
                 except Exception as e:
                     return False
 
+            print(datetime_col)
```
dorisjlee, line 471: remove this print statement

lux/executor/PandasExecutor.py
```diff
@@ -400,7 +393,7 @@ def compute_data_type(self, ldf: LuxDataFrame):
         from pandas.api.types import is_datetime64_any_dtype as is_datetime
 
         for attr in list(ldf.columns):
-            temporal_var_list = ["month", "year", "day", "date", "time"]
+            temporal_var_list = ["month", "year", "day", "date", "time", "weekday"]
```
dorisjlee, line 396: Is there a reason why we are adding "weekday" here?

dorisjlee:
Thanks @micahtyong, this looks great! Congrats on your first contribution to Lux!

PR #233, opened 2021-01-16
Add config parameters for Pandas Fallback

westernguy2:
This pull request would add a simple way to disable the catch block that lets Lux fall back to Pandas in the case of an error while running Lux.

Specifically, one would set `lux.config.pandas_fallback` to False to stop catching errors raised during `_repr_html_`. One would set `lux.config.interestingness_fallback` to False to stop catching errors raised during interestingness calculations.

These config parameters are meant for internal dev use and testing.


dorisjlee:
Thanks @westernguy2 !

PR #239, opened 2021-01-22
Update plot_config to plotting_style and documentation

caitlynachen:
Changed the 'plot_config' parameter in the lux.config class to 'plotting_style' to be in alignment with other plotting parameter, 'plotting_backend'. 
Updated the documentation accordingly


doc/source/guide/style.rst
```diff
@@ -45,18 +46,19 @@ We now see that the displayed visualizations adopt these new imported settings.
...
 Similarly, we can change the plot configurations for Matplotlib charts as well.
-The plot_config attribute for Matplotlib charts takes in both the figure and axis as parameters.
+The plotting_style attribute for Matplotlib charts takes in both the figure and axis as parameters.
+
 .. code-block:: python
 	
-	lux.config.plotting_backend = "matplotlib" # or 'matplotlib_code'
+    lux.config.plotting_backend = "matplotlib" # or 'matplotlib_code'
 
     def add_title(fig, ax):
```
dorisjlee, line 55: Can we show the same example as above of changing color and adding title. It might be good to demonstrate how we can change settings for `fig` and `ax` parameter and when we should use each. Also would we be able to set something like `plt.rcParam`? A few examples outlining the possible style changes here with links to matplotlib docs would be helpful here.

doc/source/guide/style.rst
```diff
@@ -4,7 +4,7 @@ Styling Custom Plot Settings
 
 .. note:: You can follow along this tutorial in a Jupyter notebook. [`Github <https://github.com/lux-org/lux-binder/blob/master/tutorial/4-chart-settings.ipynb>`_] [`Binder <https://mybinder.org/v2/gh/lux-org/lux-binder/master?urlpath=tree/tutorial/4-chart-settings.ipynb>`_]
 
-In the last tutorial, we saw how :code:`Vis` objects could be exported into visualization code for further editing. What if we want to change the chart settings for *all* the visualizations displayed in the widget. In Lux, we can change the chart settings and aesthetics by inputting global custom plot settings the :code:`plot_config`.
+In the last tutorial, we saw how :code:`Vis` objects could be exported into visualization code for further editing. What if we want to change the chart settings for *all* the visualizations displayed in the widget. In Lux, we can change the chart settings and aesthetics by inputting global custom plot settings the :code:`plotting_style`.
```
dorisjlee, line 8: see if we can replace `lux.config.plotting_style` with [linked domain references](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html).

doc/source/guide/style.rst
```diff
@@ -45,18 +46,19 @@ We now see that the displayed visualizations adopt these new imported settings.
   :align: center 
 
 Similarly, we can change the plot configurations for Matplotlib charts as well.
-The plot_config attribute for Matplotlib charts takes in both the figure and axis as parameters.
+The plotting_style attribute for Matplotlib charts takes in both the figure and axis as parameters.
```
dorisjlee, line 49: Why take in both figure and axis? instead of just one of these? The example below should motivate this if they are both necessary.

dorisjlee:
Thanks @caitlynachen! Let's add a test case when `plotting_style` is set for the matplotlib and reflect the tutorial changes in lux-binder.

doc/source/guide/style.rst
caitlynachen: figure can set attributes like figsize, while axis doesn't have the capabilities. axes can set attributes like plot titles

jerrysong1324 (approved):

Issue #247, opened 2021-01-27
`df.convert_dtypes()` raising new error in Pandas 1.2.0


westernguy2:
After updating to Pandas 1.2.0, the below test in `test_pandas` is now failing

```python3
df = pytest.college_df
cdf = df.convert_dtypes()
cdf._repr_html_()
assert list(cdf.recommendation.keys()) == ["Correlation", "Distribution", "Occurrence"]
```

It raises `TypeError: data type not understood` when the `cdf` is displayed with `_repr_html_`.

This test works in Pandas 1.1 but fails in Pandas 1.2 .

In the documentation for [`convert_dtypes`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.convert_dtypes.html), we see:

> Changed in version 1.2: Starting with pandas 1.2, this method also converts float columns to the nullable floating extension type.

This new change could be potentially causing the bug.

westernguy2:
In Pandas 1.2.0, a new data type `Float64Dtype` was added. It was meant to be a dtype for a column of floats, but with some null values, represented by `pd.NA`. More information about this is in the [documentation here](https://pandas.pydata.org/docs/dev/whatsnew/v1.2.0.html#experimental-nullable-data-types-for-float-data). 

Currently, when we run `convert_dtypes` in the above college DataFrame, we are converting some of the columns to `Float64Dtype`. Then, when Altair runs `sanitize_dataframe`, it calls `np.dtype` on this new data type. Unfortunately, this is what causes the `data type not understood` error. 

Currently, Altair has an [open issue](https://github.com/altair-viz/altair/issues/1931) for a similar issue with a `Period[M]` type for `sanitize_dataframe`.

We can either sanitize the data ourselves before we send it to Altair, or we'll have to wait for a patch from likely Altair.

dorisjlee:
Thanks for looking into this @westernguy2! 

Let's sanitize the data ourselves before sending it to Altair by converting the `Flaot64Dtype` to something that Altair can understand. We can do this inside [AltairChart:sanitize_dataframe](https://github.com/lux-org/lux/blob/3393b9ff5c240c624250a900f4aa68ec7caa994c/lux/vislib/altair/AltairChart.py#L118) similar to how we handle column names that can not be visualized by Altair. 

Could we also check that this `Float64Dtype` problem does not cause an issue for matplotlib?

dorisjlee:
[...]

PR #250, opened 2021-02-01
Fix nullable dtypes not compatible with Altair

westernguy2:
This fixes the issue in #247. It checks if the dtype is any float value, including `Float64Dtype` and converts it to the numpy float, which is compatible with Altair.


dorisjlee:
Thanks @westernguy2 ! This is now working with Pandas 1.2.1.

PR #253, opened 2021-02-06
New Data Type + Action: Automatic geographic attribute visualizations 

### LINKED ISSUE #123: New Action: Automatic geographic attribute visualizations
For geo-related attributes (longitude, latitude, state, countries, etc), we should detect it as a special data type and display a choropleth automatically.

micahtyong:
## In this PR

Closes #123 by adding a new data type, `geographical`, which detects columns representing geographic attributes (e.g. `state`, `country`), then calls the default action, `univariate`, which then aggregates quantitative attributes and displays the corresponding [choropleth symbol map](https://en.wikipedia.org/wiki/Choropleth_map). 

## Changes
- Addition of new `geographical` datatype
- Modified the `univariate` default action and added `SymbolMap` visualization
- Utilizes existing aggregations from `PandasExecutor.py`
- Tests in `tests/test_type.py`, `tests/test_action.py`, and `test_vegalite_default_actions_registered` for correctness and failure cases

## Example Output
Screen Shot 2021-02-20 at 6 03 58 PM


micahtyong:
cc: @dorisjlee. Just worked through [this issue](https://github.com/lux-org/lux/issues/123)! Lux can now detect and visualize geographical attributes. I may have pulled upstream changes in a way s.t. this PR includes merged commits from other PRs, but the logic from those remains unchanged. Let me know if there are changes I might need to make. I hope this helps!

lux/executor/Executor.py
```diff
@@ -57,7 +57,7 @@ def compute_data_type(self):
 
     def mapping(self, rmap):
         group_map = {}
-        for val in ["quantitative", "id", "nominal", "temporal"]:
+        for val in ["quantitative", "id", "nominal", "temporal", "geoshape"]:
```
dorisjlee, line 60: Can we rename the data type name from`geoshape` to `geographical`?

lux/executor/Executor.py
```diff
@@ -74,7 +74,7 @@ def invert_data_type(self, data_type):
     def compute_data_model(self, data_type):
         data_type_inverted = self.invert_data_type(data_type)
         data_model = {
-            "measure": data_type_inverted["quantitative"],
+            "measure": data_type_inverted["quantitative"] + data_type_inverted["geoshape"],
```
dorisjlee, line 77: In the current system, secondary geo attributes like `state` are still declared as nominal. I wonder if we should be labeling them as geo attributes also. If so this may need to account for non-quantitative geo types?
For now, let's take out the geo data type from measures so that they are not plotting in the correlation tab in the example shown.

tests/test_type.py
```diff
@@ -100,6 +100,22 @@ def test_check_airbnb():
     }
 
 
+def test_check_airports():
+    from vega_datasets import data
```
dorisjlee, line 104: `vega_datasets` is not a test dependency. We can just link to the dataset directly: 
```python
df = pd.read_csv('https://raw.githubusercontent.com/altair-viz/vega_datasets/master/vega_datasets/_data/airports.csv')
```

lux/action/map.py
```diff
@@ -0,0 +1,108 @@
...
+        "description": "Show proportional symbol maps of <p class='highlight-descriptor'>geographic</p>  attributes.",
+    }
+
+    if len(ldf) < 5 or len(possible_attributes) < 2:
+        ignore_rec_flag = True
+    if ignore_rec_flag or not valid_geoshape(possible_attributes):
+        recommendation["collection"] = []
+        return recommendation
+
+    intent = [lux.Clause("?", data_model="measure"), lux.Clause("?", data_model="measure")]
```
dorisjlee, line 57: Would other measure variables be also plotted based on this intent?

dorisjlee (changes requested):

tests/test_type.py
micahtyong: Gotcha, thanks for the tip!

lux/action/map.py
micahtyong: No, since the `geomap` action specifically looks for attributes that are typed `geographical`.

lux/executor/Executor.py
micahtyong: I see. I may have to take a closer look at this and ask you about it during our 1:1 since geographical attributes being recognized depend on it being part of the measures.

tests/test_vis.py
```diff
@@ -463,6 +482,7 @@ def test_vegalite_heatmap_flag_config():
     df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/airbnb_nyc.csv")
     df = df.copy()
     assert not df.recommendation["Correlation"][0]._postbin
+    assert "Geographic" not in df.recommendation
```
caitlynachen, line 485: have you looked into Matplotlib supported Geographic charts?

caitlynachen: just wondering for future PRs:)

micahtyong: Hi @caitlynachen! Yes (: Matplotlib has an API called [basemap](https://matplotlib.org/basemap/) which allows you to make geographic map projections just like the one I did in Altair. There's a good example of the proportional symbol map in [this article](https://jakevdp.github.io/PythonDataScienceHandbook/04.13-geographic-data-with-basemap.html). Let me know what you think!

lux/interestingness/interestingness.py
```diff
@@ -363,3 +364,9 @@ def monotonicity(vis: Vis, attr_specs: list, ignore_identity: bool = True) -> in
         return -1
     else:
         return score
+
+
+def n_distinct(vis: Vis, dimension_lst: list, measure_lst: list) -> int:
```
dorisjlee, line 369: What is the goal of the `n_distinct`? I didn't quite understand this and was wondering if we should just assign dummy scores, since it is unlikely that a dataset would have too many geo attributes. Similar to what we do in `generalize.py`.

lux/vislib/altair/SymbolMap.py
```diff
@@ -0,0 +1,164 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from lux.vislib.altair.AltairChart import AltairChart
+import altair as alt
+import us
+import pycountry
```
dorisjlee, line 18: We should list us and pycountry into the requirement.txt. 
I installed these via pip and noticed that pycountry in particular is quite a big installation (10MB). Perhaps we could just use a dictionary for doing this lookup to keep the package minimal?

lux/vislib/altair/SymbolMap.py
```diff
@@ -0,0 +1,164 @@
...
+
+alt.data_transformers.disable_max_rows()
+
+
+class SymbolMap(AltairChart):
+    """
+    SymbolMap is a subclass of AltairChart that renders choropleth maps.
```
dorisjlee, line 25: Should we rename this GeoHeatmap or Choropleth now that we replaced SymbolMap with the heatmaps?

lux/utils/utils.py
```diff
@@ -113,6 +113,10 @@ def like_nan(val):
         return math.isnan(val)
 
 
+def like_geo(val):
+    return isinstance(val, str) and val.lower() in {"state", "country"}
```
dorisjlee, line 117: This is a fine implementation for now. Perhaps in the next PR, we can think about making longitude and latitude a geo type also but somehow labelled differently than `state` and `country`. Since that way longitude and latitude would not get lumped with other quantitative types, to be displayed in `Correlation`. 

image

dorisjlee (changes requested):

lux/interestingness/interestingness.py
micahtyong: `n_distinct` computes how many unique values there are for a geographical data type. For example, if a dataset displayed earthquake magnitudes across 48 states and 3 countries, `n_distinct` would return 48 and 3 respectively, meaning the states choropleth would be displayed first, and the countries choropleth second.

dorisjlee: Ah ok, this makes sense!

lux/vislib/altair/SymbolMap.py
micahtyong: Sounds good, I've swapped `pycountry` out with [iso3166](https://pypi.org/project/iso3166/#description), which enables me to do the same thing with only 36 kb (much much lighter)

lux/vislib/altair/SymbolMap.py
micahtyong: Yup, changed this in the latest commit!

micahtyong:
Hi @dorisjlee, I've added the ability to export choropleths, as requested. An example output is shown below. Let me know if there's anything else I should modify! 

Screen Shot 2021-02-25 at 1 13 50 AM

dorisjlee:
Hi @micahtyong, 
This looks really great! I tried this out on the Covid dataset that we worked on a while back and the maps look really cool with the different measure values in the dataset. 
```python
df['dateBefore'] = pd.to_datetime(df['dateBefore'], format='%Y-%M-%d')
df
```
image
One thing that I did notice from this dataset is that when we click on the visualization, there is a lot of "jitter" that happen with the plots being redrawn. 
Feb-27-2021 21-49-24
I believe this is related to [lux-widget:#63](https://github.com/lux-org/lux-widget/issues/63). This re-drawing issue is present for all charts, but perhaps because the map vis takes longer to draw, so the jitter looks more off. This is a bigger issue that should be addressed separately, so I think we should be good to merge this issue independent of this.

I'll wait until @caitlynachen does a code review on this PR before merging this in. 
Thanks!

dorisjlee (approved):

lux/interestingness/interestingness.py
```diff
@@ -112,7 +114,6 @@ def interestingness(vis: Vis, ldf: LuxDataFrame) -> int:
             if v_size < 10:
                 return -1
             color_attr = vis.get_attr_by_channel("color")[0].attribute
-
```
caitlynachen, line 115: remove change?

caitlynachen (approved):
Looks great! Maybe next steps is to work on the same geo feature for Matplotlib, but amazing job figuring it out for Altair!

lux/vislib/altair/Choropleth.py
```diff
@@ -0,0 +1,180 @@
...
+
+        chart = background + points
+
+        ######################################
+        ## Constructing Altair Code String ##
+        #####################################
+
+        self.code += "import altair as alt\n"
+        dfname = "placeholder_variable"
+        self.code += f"""nan=float('nan')
```
caitlynachen, line 89: Is this formatting expected? Make sure it passes "black ."?

dorisjlee (approved):

dorisjlee:
Thanks @micahtyong!

Issue #61, opened 2020-08-13
Better data type detection for pre_aggregated, indexed dataframes


dorisjlee:
When a dataframe is pre-aggregated, our type detection based on cardinality often fail to detect the type correctly. For example, when the dataset size is small (often the case when data is pre-aggregated), nominal fields would get recognized as a `quantitative` type.
```
df = pd.read_csv("lux/data/car.csv")
df["Year"] = pd.to_datetime(df["Year"], format='%Y') # change pandas dtype for the column "Year" to datetype
a = df.groupby("Cylinders").mean()

a.data_type
```
<img src="https://user-images.githubusercontent.com/5554675/90112958-3f428600-dd83-11ea-9645-187788fb29e3.png" width=250></img>

As a related issue, we should also support the detection of types for named index, for example, in this case, `Cylinders` is an index, so its data type is not being computed.

dorisjlee:
Closing this after #287 is merged in. Great work Kunal!

PR #260, opened 2021-02-10
LuxGroupby Implementation

westernguy2:
This pull request extends the Pandas Groupby Object and creates a LuxGroupBy object. The purpose of this is to track when a `groupby` is called by the user using the history. The `history` is then propagated through the `LuxGroupBy` object, along with the rest of the metadata, and through the different aggregate or filter functions to the newly created LuxDataFrame.

Lux will use the history to see if the current LuxDataFrame is the result of a `groupby`. If so, it'll set `pre_aggregated` to True, and will show `Column Group` and `Row Group` visualizations.

Internally, all `groupby` operations we don't want to record will have an extra argument `history` which must be set to False.

This pull request also fixes an issue surrounding `get_axis_number` with metadata propagation from `LuxSeries` to `LuxDataFrame`.

This PR also fixes a bug in `LuxSeries`. When a Series is instantiated, the metadata is all set to None. So, when it moves to a newly created LuxDataFrame, some of the metadata is set to None instead of the default value such as an empty dictionary or an empty list. This is fixed by setting the default values at `__init__` if they're not None.

This PR makes significant progress towards the issues described in #61 and is the start of using user history to recommend certain visualizations.


tests/test_interestingness.py
```diff
@@ -21,7 +21,7 @@
 
 # The following test cases are labelled for vis with <Ndim, Nmsr, Nfilter>
 def test_interestingness_1_0_0(global_var):
-    df = pytest.car_df
+    df = pd.read_csv("lux/data/car.csv")
```
dorisjlee, line 24: Is there a reason why all of these were switched back to `read_csv`?

tests/test_pandas_coverage.py
```diff
@@ -595,7 +595,12 @@ def test_df_to_series(global_var):
         "_prev",
         "_history",
         "_saved_export",
```
dorisjlee, line 597: Can we share this list of variables to check so that we don't have to update all of them across the different test functions?

lux/core/groupby.py
```diff
@@ -0,0 +1,65 @@
...
+        "_message",
+        "_pandas_only",
+        "pre_aggregated",
+        "_type_override",
+    ]
+
+    def __init__(self, *args, **kwargs):
+        super(LuxDataFrameGroupBy, self).__init__(*args, **kwargs)
+
+    def aggregate(self, *args, **kwargs):
```
dorisjlee, line 31: Could you add some comments for each overridden function on the rationale on choosing to propogate metadata for some but not other (e.g. `filter`, `get_group`)? Is there a way to do this more uniformly across the functions?

dorisjlee (changes requested):

lux/executor/PandasExecutor.py
```diff
@@ -163,17 +163,20 @@ def execute_aggregate(vis: Vis, isFiltered=True):
...
                         .count()
                         .reset_index()
                         .rename(columns={index_name: "Record"})
                     )
                     vis._vis_data = vis.data[[groupby_attr.attribute, color_attr.attribute, "Record"]]
                 else:
                     vis._vis_data = (
-                        vis.data.groupby(groupby_attr.attribute, dropna=False)
+                        vis.data.groupby(groupby_attr.attribute, dropna=False, history=False)
                         .count()
```
jinimukh, line 180: Since we're not recording in the history, does that mean that the history of the vis._data is no longer correct? I.e. if vis._data contains an aggregation, shouldn't its history say that it contains an aggregation?

 I think this brings up a bigger issue that the latest event should be recorded on the latest version of the dataframe, but I think we are recording the latest history on the dataframe that is transformed to get the latest dataframe. Is this correct?

jinimukh (changes requested):
Can we add a few tests about what the expected action of History should be? Otherwise, this looks good!! Great work @westernguy2 !!

Edit: Just saw the tests! I'm changing this approved! Really cool work!

jinimukh (approved):

tests/test_interestingness.py
westernguy2: Yah, since it's the same DataFrame being used across different tests, the History is propagated through, so if one test did a groupby, all future tests will have groupby in the history of the `df`. So, for the ones with a groupby, we had to load in the data directly.

tests/test_pandas_coverage.py
westernguy2: Yes, I'll add that!

lux/executor/PandasExecutor.py
westernguy2: Not sure if I'm understanding correctly, but the history is tied to the user's DataFrame, not any DataFrames that we use internally like `vis.data`. The history is tied to both the DataFrame where the operation is acted on and also newly created DataFrames from that DataFrame. 

Now that I think about it, if I do something like

```python
df.groupby('Origin')
new_df = df.apply(f)
```
The `new_df` might have a groupby in its history, which wouldn't be correct. I'll look into this!

lux/core/groupby.py
westernguy2: Will add some comments! I believe for `filter` and `get_group` we were more concerned about making sure `pre_aggregated` was not set to True. I believe the metadata was already being propagated through, but I'm not sure why so I'll look into this!

westernguy2:
> Can we add a few tests about what the expected action of History should be?

Sure, will do! Thanks!

lux/executor/PandasExecutor.py
westernguy2: I moved the history out of the original df, so it'll only exist in the groupby object and the DataFrame resulting from the groupby operation.

lux/core/groupby.py
westernguy2: Added comments! I realized I was definitely supposed to be propagating the metadata through `filter` and `get_group`, so I've added that. Thanks for the catch!

tests/test_pandas_coverage.py
westernguy2: Set a variable in `conftest` for the metadata list, so we just have to edit that list.

dorisjlee (approved):

dorisjlee:
This is really great work, thanks @westernguy2!

Issue #243, opened 2021-01-25
Improve Temporal action to display different time scales


dorisjlee:
For temporal attributes like dates, the column can often be visualized at different time-scales. For example, a date formatted as %y-%m-%d can be plotted as 3+1 separate visualizations: one based on year, one based on the month, and one based on day, and one based on the overall date. We should extend the temporal action to display the temporal columns at these different time scales.

dorisjlee:
Merged #262

PR #262, opened 2021-02-11
Adding support to improve temporal action to display different timescales

mantejpanesar:
### In this PR
This PR addresses #243 by adding a separate `temporal` action, which generates visualizations at different timescales for temporal columns. Supports separate visualizations for the following formats: `%Y-%m-%d`, `%Y-%m`, and single component temporal columns (`%Y`, `%m`, `%d`).

### Changes
* Add a new `temporal` action to handle the creation of visualizations for temporal related columns.
* Tests in `test/test_action.py` to verify that the correct type and number of visualizations are being generated.

### Visualizations
The following images showcases the new functionality for visualizing temporal columns tested on the Airbnb dataset.

#### Temporal Visualization Before
Screen Shot 2021-02-11 at 12 29 55 PM

#### Temporal Visualization After
Screen Shot 2021-02-11 at 12 21 29 PM

### DIFF
--- a/lux/action/default.py
+++ b/lux/action/default.py
@@ -6,6 +6,7 @@ def register_default_actions():
     from lux.action.enhance import enhance
     from lux.action.filter import add_filter
     from lux.action.generalize import generalize
+    from lux.action.temporal import temporal
 
     # display conditions for default actions
     no_vis = lambda ldf: (ldf.current_vis is None) or (
@@ -18,7 +19,7 @@ def register_default_actions():
     lux.config.register_action("correlation", correlation, no_vis)
     lux.config.register_action("distribution", univariate, no_vis, "quantitative")
     lux.config.register_action("occurrence", univariate, no_vis, "nominal")
-    lux.config.register_action("temporal", univariate, no_vis, "temporal")
+    lux.config.register_action("temporal", temporal, no_vis)
     lux.config.register_action("geographical", univariate, no_vis, "geographical")
 
     lux.config.register_action("Enhance", enhance, one_current_vis)
--- a/lux/action/temporal.py
+++ b/lux/action/temporal.py
@@ -17,7 +17,6 @@ from lux.vis.VisList import VisList
 from lux.vis.Vis import Vis
 import pandas as pd
 from lux.core.frame import LuxDataFrame
-from lux.utils.date_utils import compute_date_granularity
 from lux.interestingness.interestingness import interestingness
 from lux.utils import utils
 
@@ -34,33 +33,32 @@ def temporal(ldf):
     recommendations : Dict[str,obj]
             Object with a collection of visualizations that result from the Temporal action.
     """
-    visuals = []
+    vlist = []
     recommendation = {
         "action": "Temporal",
         "description": "Show trends over <p class='highlight-descriptor'>time-related</p> attributes.",
         "long_description": "Temporal displays line charts for all attributes related to datetimes in the dataframe.",
     }
-
     for c in ldf.columns:
         if ldf.data_type[c] == "temporal":
             try:
                 generated_vis = create_temporal_vis(ldf, c)
-                visuals.extend(generated_vis)
-            except (ValueError, AttributeError, KeyError):
+                vlist.extend(generated_vis)
+            except:
                 pass
 
     # If no temporal visualizations were generated via parsing datetime, fallback to default behavior.
-    if len(visuals) == 0:
+    if len(vlist) == 0:
         intent = [lux.Clause("?", data_type="temporal")]
         intent.extend(utils.get_filter_specs(ldf._intent))
         vlist = VisList(intent, ldf)
         for vis in vlist:
             vis.score = interestingness(vis, ldf)
     else:
-        vlist = VisList(visuals)
+        vlist = VisList(vlist)
         recommendation["long_description"] += (
-            " The scoring will display all overall visualizations first,"
-            + "all the (year) visualizations second, all the (month) visualizations third, and finally all the (day) visualizations."
+            " Lux displays the overall temporal trend first,"
+            + " followed by trends across other timescales (e.g., year, month, week, day)."
         )
 
     # Doesn't make sense to generate a line chart if there is less than 3 datapoints (pre-aggregated)
@@ -74,7 +72,7 @@ def temporal(ldf):
 
 def create_temporal_vis(ldf, col):
     """
-    Creates and populates Vis objects for the different time-scale visualizations of the provided temporal column.
+    Creates and populates Vis objects for different timescales in the provided temporal column.
     Parameters
     ----------
     ldf : lux.core.frame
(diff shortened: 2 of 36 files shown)


lux/action/temporal.py
```diff
@@ -0,0 +1,109 @@
...
+    if len(ldf) < 3:
+        recommendation["collection"] = []
+        return recommendation
+    vlist = VisList(visuals)
+    vlist.sort()
+    recommendation["collection"] = vlist
+    return recommendation
+
+
+def create_vis(ldf, col):
```
dorisjlee, line 58: Rename `create_vis` to be specific to temporal, since it is easily mistaken as the other `create_vis` inside the renderer

lux/action/temporal.py
```diff
@@ -0,0 +1,109 @@
...
+            Name of temporal column.
+
+    Returns
+    -------
+    visuals : [Vis]
+            Collection of Vis objects.
+    """
+    visuals = []
+    converted_string = ldf[col].astype(str)
+    if converted_string.str.contains("-").any():
```
dorisjlee, line 77: I think this can be done without the regex string by simply [extracting the timescale values from the datetime attribute via Pandas](https://stackoverflow.com/questions/25146121/extracting-just-month-and-year-separately-from-pandas-datetime-column), such as `.dt.year`, `.dt.month`, etc. We should avoid having to do `astype` since this can be an expensive operation.

tests/test_action.py
```diff
@@ -19,6 +19,56 @@
...
+                "2020-05",
+                "2020-01",
+                "2020-02",
+                "2020-03",
+                "2020-04",
+                "2020-05",
+            ]
+        }
+    )
+    test_data = [airbnb_df, flights_df, date_df, pytest.car_df, pytest.olympic]
```
dorisjlee, line 55: For the flights dataset, the results look a bit weird since the different timescales are created even for attributes that only encompass a single timescale.
```python
df = pd.read_csv("../../lux-datasets/data/flights.csv")
df['year'] = pd.to_datetime(df['year'], format='%Y')
df['month'] = pd.to_datetime(df['month'], format='%M')
df['day'] = pd.to_datetime(df['day'], format='%d')
```
image

lux/action/temporal.py
```diff
@@ -0,0 +1,109 @@
...
+        parsed_date = converted_string.str.extract(r"([0-9]{4})?-?([0-9]{2})?-?([0-9]{1,2})?")
+        valid_year, valid_month, valid_day = parsed_date.apply(lambda x: not x.isnull().all()).values
+
+        date_vis = Vis([lux.Clause(col, data_type="temporal")], source=ldf, score=4)
```
dorisjlee, line 81: We should discard time scales that are not applicable, for example a day will show up as all `*-*-01` if it only has year and month information
```python
df = pd.read_csv("../../lux-datasets/data/stocks.csv")
df
```
image

lux/action/temporal.py
```diff
@@ -0,0 +1,109 @@
...
+            visuals.append(month_vis)
+        if valid_day:
+            day_col = col + " (day)"
+            day_df = LuxDataFrame(
+                {day_col: pd.to_datetime(parsed_date[2], format="%d").dt.strftime("%d")}
+            )
+            day_vis = Vis([lux.Clause(day_col, data_type="temporal")], source=day_df, score=1)
+            visuals.append(day_vis)
+    else:
+        single_vis = Vis([lux.Clause(col, data_type="temporal")], source=ldf, score=5)
```
dorisjlee, line 107: How does this scoring behave when there are more than two temporal attributes? Do we show the unseparated overall first, then move onto all `(year)`, `(month )`, etc? Perhaps add an example of this in the use cases.

dorisjlee (changes requested):

dorisjlee:
Thanks @mantejpanesar! This looks really great as a first issue, I left some comments in the code review for you to address.

lux/action/default.py
```diff
@@ -19,8 +19,7 @@ def register_default_actions():
     lux.config.register_action("correlation", correlation, no_vis)
     lux.config.register_action("distribution", univariate, no_vis, "quantitative")
     lux.config.register_action("occurrence", univariate, no_vis, "nominal")
-    lux.config.register_action("temporal", univariate, no_vis, "temporal")
-    lux.config.register_action("geographical", geomap, no_vis, "geographical")
```
jerrysong1324, line 23: @micahtyong I'm not sure if you are depending on this, but if so, let's not remove this import.

lux/action/temporal.py
```diff
@@ -0,0 +1,109 @@
...
+    for c in ldf.columns:
+        if ldf.data_type[c] == "temporal":
+            try:
+                generated_vis = create_vis(ldf, c)
+                visuals.extend(generated_vis)
+            except (ValueError, AttributeError):
+                pass
+    recommendation = {
+        "action": "Temporal",
+        "description": "Show trends over <p class='highlight-descriptor'>time-related</p> attributes.",
```
jerrysong1324, line 46: Want to try adding in a long description field as well? Our long descriptions include the description and also how the graphs are sorted and other additional info for the user.

lux/action/temporal.py
jerrysong1324: The answer to this can be included into the long description.

lux/action/default.py
micahtyong: Yeah, I am, thanks for the catch! @jerrysong1324

lux/action/temporal.py
mantejpanesar: Thanks for the catch! I'll update the function name.

lux/action/temporal.py
mantejpanesar: Sounds good, I'll update the action to directly extract the different timescales from the datetime attribute.

tests/test_action.py
mantejpanesar: I'll take a closer look at these results. The fixes/changes to utilize the datetime attribute and discarding timescales that are not applicable should end up resolving this issue. I'll make the changes and ensure that these visualizations are proper.

lux/action/temporal.py
mantejpanesar: Sounds good, I'll update the logic in the temporal action to use `compute_date_granularity()` in order to discard time scales that aren't applicable.

lux/action/default.py
mantejpanesar: Thanks for the catch!

lux/action/temporal.py
mantejpanesar: The scoring will result in the overall visualizations first, then all the `(year)` visualizations, all the `(month)` visualizations, etc. Is there an alternate ordering that should be considered? I'll also add an example for this situation to the testing and use cases and update the long description with the scoring behavior.

lux/action/temporal.py
mantejpanesar: Thanks for clarifying what the long descriptions are used for. I'll add how the graphs are sorted in the long description.

mantejpanesar:
Hi @dorisjlee, As requested, I've updated the temporal action to parse the temporal fields using the datetime attribute via Pandas and also handled discarding timescales that are not applicable. I have provided an example output for the stocks dataset, the flights dataset, and an example output using fabricated data (check-in and check-out dates) to highlight the scoring behavior for multiple temporal columns with multiple timescales. Let me know if there's any other changes I should make!

#### Stocks Dataset Visualization
This demonstrates that timescales which are not applicable are no longer presented in the visualizations.
StocksDataset

#### Flights Dataset Visualization
This demonstrates the removal of erroneous visualizations for attributes that only encompass a single timescale.
FlightsDataset

#### Scoring Behavior
This demonstrates the scoring behavior for the temporal action. When visualizations are generated via parsing a temporal column with the datetime attribute, all the overall visualizations are shown first, followed by all the (year) visualizations, followed by all the (month) visualizations, and finally all the (day) visualizations.
ScoringBehavior

lux/action/temporal.py
micahtyong: Just popping in (: In my opinion, the current ordering can probably be improved. We have a suite of functions in `interestingness.py` that give you a sense of how visualizations are ranked. Perhaps `weighted_correlation` is a good place to start? This doesn't have to be in this PR, but just something to think about.

tests/test_action.py
micahtyong: I noticed in your updated [results](https://github.com/lux-org/lux/pull/262#issuecomment-787717020) that the range for the y-axis doesn't fit the data well (e.g. the minimum of the data points are far higher than the minimum of the y-range). Do you know why that might be?

lux/action/temporal.py
```diff
@@ -0,0 +1,133 @@
...
+    year_df = LuxDataFrame({year_col: pd.to_datetime(formatted_date.dt.year, format="%Y")})
+    year_vis = Vis([lux.Clause(year_col, data_type="temporal")], source=year_df, score=3)
+
+    month_col = col + " (month)"
+    month_df = LuxDataFrame({month_col: pd.to_datetime(formatted_date.dt.month, format="%m")})
+    month_vis = Vis([lux.Clause(month_col, data_type="temporal")], source=month_df, score=2)
+
+    day_col = col + " (day)"
+    day_df = LuxDataFrame({day_col: pd.to_datetime(formatted_date.dt.day, format="%d")})
+    day_vis = Vis([lux.Clause(day_col, data_type="temporal")], source=day_df, score=1)
```
micahtyong, line 108: I find it odd that we're building the data frame and vis directly from the action class rather than using classes like `VisList` and `PandasExecutor`. At the same time, I see where you were coming from (we're essentially extending making copies of the same data frame just using different time scales). I'll leave my comment here in case Doris has any input on that!

micahtyong (commented):
Hi @mantejpanesar, just stopping by. I left a few comments on this PR, but overall, the results look great for a first issue (adding or modifying data types can be tricky!).

lux/action/temporal.py
mantejpanesar: Hi @micahtyong, thanks for the review! I'm not sure why the y-axis seems to have a minimum far below the data points. The fix in this PR relies on the existing functionality to populate the visualizations; it appears that the same y-axis range is being produced even when the changes from this PR are not added -- for instance, the flights dataset -- but I can take a closer look at it!

mantejpanesar:
Hi @micahtyong, thanks for the review! I'm not sure why the y-axis seems to have a minimum far below the data points. The fix in this PR relies on the existing functionality to populate the visualizations; it appears that the same y-axis range is being produced even when the changes from this PR are not added -- for instance, the flights dataset -- but I can take a closer look at it!

dorisjlee:
Thanks @mantejpanesar, @micahtyong! I have made some final fixes and tests for this feature. A future to-do would be to extend the temporal action to support the different timescales on temporal attributes more generically, beyond just the Record-based line charts, but also when a measure value is present (e.g., in the case of Enhance). This would make sense for use cases like [this](https://altair-viz.github.io/user_guide/transform/timeunit.html), where the measure is not counts. But this is great as a first start!

PR #264, opened 2021-02-14
Conda setup 

cjachekang:
Finished all of the work on the conda-forge side - merging in documentation for how to install via Conda and the yaml files so that we can update the recipe in the future.

In the future, in order to update the recipes, you can follow the directions here: https://conda-forge.org/#update_recipe
We should only have to update the recipes for new dependencies/updating version numbers. 

You can install via `conda install -c conda-forge lux-api`


conda.recipe/lux-api/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "lux-api" %}
+{% set version = "0.2.2" %}
```
dorisjlee, line 2: Is it possible to read this number directly from [_version.py](https://github.com/lux-org/lux/blob/master/lux/_version.py) so that we wouldn't have to update the recipe every time there is a new release? A similar thing that reads from Github tags is done [here](https://github.com/modin-project/modin/blob/master/scripts/conda-recipe/meta.yaml)

conda.recipe/lux-widget/meta.yaml
```diff
@@ -0,0 +1,48 @@
+{% set name = "lux-widget" %}
```
dorisjlee, line 1: Should we put this file inside the lux-widget repo?

conda.recipe/lux-widget/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+{% set version = "0.1.3" %}
```
dorisjlee, line 2: Similiarly, this way the version can be read in from [_version.py](https://github.com/lux-org/lux-widget/blob/master/luxwidget/_version.py) or package.json

dorisjlee (approved):

dorisjlee:
Thanks @cjachekang! I tested this on [Docker with Anaconda](https://hub.docker.com/r/continuumio/anaconda3) and it works perfectly. I made minor adjustment to make the README install instruction a one-liner and left some comments. Could you add some description of when and how to update the conda recipes, including links and which files to modify inside CONTRIBUTING.md?

conda.recipe/lux-widget/meta.yaml
cjachekang: Yeah I can do that

conda.recipe/lux-api/meta.yaml
cjachekang: I can try this approach out, I have already uploaded 0.2.2 to conda-forge so I will try it for the new release

dorisjlee: Ok that sounds good!

dorisjlee:
Thanks @cjachekang !

Issue #259, opened 2021-02-09
Export functionality not supported in LuxSeries


dorisjlee:
Currently, the exported property is not supported in LuxSeries, we need to extend LuxSeries to export the visualization generated as LuxSeries using similar logic as what is in [frame.py](https://github.com/lux-org/lux/blob/master/lux/core/frame.py#L506).

PR #277, opened 2021-02-19
Add export Vis option to LuxSeries

caitlynachen:
https://github.com/lux-org/lux/issues/259
Saves widget in repr, in order to export selected indices in exported function
Screen Shot 2021-02-19 at 3 19 28 PM


lux/core/series.py
```diff
@@ -211,7 +215,83 @@ def recommendation(self):
             self.name = " "
         ldf = LuxDataFrame(self)
 
-        if self._recommendation is not None and self._recommendation == {}:
```
dorisjlee, line 214: why is this if statement taken out?

lux/core/series.py
```diff
@@ -211,7 +215,83 @@ def recommendation(self):
...
-            ldf.maintain_metadata()
-            ldf.maintain_recs()
+        ldf.maintain_metadata()
+        ldf.maintain_recs()
         return ldf._recommendation
+
+    @property
+    def exported(self) -> Union[Dict[str, VisList], VisList]:
```
dorisjlee, line 223: Can we reuse and share this code with the `exported` property function in `frame.py`?

lux/core/series.py
```diff
@@ -211,7 +215,83 @@ def recommendation(self):
             self.name = " "
         ldf = LuxDataFrame(self)
```
dorisjlee, line 216: Similar to what we did in frame.py, can we cache the result of the maintain_metadata and recs so that we do not have to create a new LuxDataFrame and regenerate the recommendations every time?

dorisjlee (changes requested):

dorisjlee:
Thanks @caitlynachen, this looks like a good start! I think there are some design decisions around where code can be reused across frame.py and series.py. In addition, we should ensure that the recommendations are not recomputed following the same caching logic as LuxDataFrame. Perhaps you can discuss with @westernguy2 on how to proceed?

lux/core/series.py
caitlynachen: Recommendation is never set in series.py, and in frame.py the recommendation is always recomputed, just in case recommendation is called without calling repr_html first

lux/core/series.py
caitlynachen: yup fixed now!

lux/core/series.py
caitlynachen: actually fixed now!

lux/core/series.py
caitlynachen: got it!

dorisjlee (approved):

dorisjlee:
Thanks @caitlynachen! The new implementation is much cleaner, merging this in now!

PR #279, opened 2021-02-20
Display better warnings when no recommendations are generated

jinimukh:
This PR allows the backend to send messages to the frontend API to be displayed as warnings rather than using UserWarnings which do not comply with Pandas Style.
So far, the only case when no recommendations are shown is when none of the default actions can be completed because the data has too few data points.

## Changed
- actions only computed if length of df > 0
- `_data_type` set to {} instead of `None` to enable fetching from an empty dictionary
- appropriate error messages added
- removed Pandas fallback
- adjust some series condition checks

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -429,7 +429,7 @@ class LuxDataFrame(pd.DataFrame):
         if recommendations["collection"] is not None and len(recommendations["collection"]) > 0:
             rec_infolist.append(recommendations)
 
-    def maintain_recs(self, is_series=False):
+    def maintain_recs(self, is_series="DataFrame"):
         # `rec_df` is the dataframe to generate the recommendations on
         # check to see if globally defined actions have been registered/removed
         if lux.config.update_actions["flag"] == True:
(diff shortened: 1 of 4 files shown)


lux/core/frame.py
```diff
@@ -451,17 +448,26 @@ def maintain_recs(self):
...
-                id_fields_str += f"<code>{id_field}</code>, "
-            id_fields_str = id_fields_str[:-2]
-            rec_df._message.add(f"{id_fields_str} is not visualized since it resembles an ID field.")
+        if len(rec_df) == 0:
+            rec_df._message.add(
+                f"Lux cannot operate on empty DataFrames or Series; Lux requires at least 5 data points to suggest visualizations."
+            )
+        elif len(rec_df) < 5:
+            rec_df._message.add(
+                f"Lux could not compute any actions because the DataFrame or Series is too small (less than 5 data points)."
```
dorisjlee, line 457: Same comment as above on picking {DataFrame/Series}.

Minor phrase change: 
```
f"The {DataFrame} is too small to visualize. To generate visualizations in Lux, the {DataFrame} must contain at least 5 rows."
```

lux/core/series.py
```diff
@@ -138,14 +140,13 @@ def __repr__(self):
...
                     return ""
-
-                if len(self) <= 0:
-                    warnings.warn(
-                        "\nLux can not operate on an empty series.\nPlease check your input again.\n",
-                        stacklevel=2,
-                    )
-                    print(series_repr)
-                    return ""
+                # if len(self) <= 0:
```
dorisjlee, line 143: We can remove the comment altogether.

For some reason, I'm still seeing the Pandas warning, is this normal?
image

dorisjlee (changes requested):

lux/core/series.py
jerrysong1324: I see this as well.

jinimukh: Resolved, I missed a push -- I'll update y'all when I push again!

dorisjlee (approved):

PR #281, opened 2021-02-21
[PATCH] Require Pandas 1.2.0

jinimukh:
This PR updates the requirement for the Pandas package to 1.2.0. Otherwise, it causes an unexpected error in `pd.crosstab` and does not accurately create the right visualizations.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 scipy>=1.3.3
 altair>=4.0.0
 numpy>=1.16.5
-pandas>=1.1.0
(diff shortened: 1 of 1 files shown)


dorisjlee:
Thanks for catching this @jinimukh! This is probably related to the new changes in pandas 1.2.0, in particular [changes related to reshaping](https://pandas.pydata.org/docs/whatsnew/v1.2.0.html#reshaping). Increasing the pandas min version should solve the issue.

PR #282, opened 2021-02-21
Changing Lux interaction from button to tab

cjachekang:
This leverages ipywidgets Tab to load lux in the background while the pandas view is initially displayed. Tabs would also prevent the widget from re-rendering everytime we switched between the pandas dataframe and the widget.

One thing to note is I am failing 3 tests (even on the master branch). This is normal right? I don't see travis being ran.


cjachekang:
[Closing in favor of full lazy streaming implementation]

Issue #217, opened 2021-01-11
Improve automatic bin determination for histograms


dorisjlee:
Currently, the formula for histogram binning sometimes results in bins that are very "skinny" and sometimes bins that are very "wide". We need to improve histogram bin width and size determination to ensure more accurate histograms are plotted. 
This is especially true for the "Filter" action.

Example: 
```python
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/olympic.csv?raw=True")
df.intent=["Height"]
df
```
![image](https://user-images.githubusercontent.com/5554675/104189469-aa3cd980-5455-11eb-94c6-06849b836e5a.png)

![image](https://user-images.githubusercontent.com/5554675/104189512-bd4fa980-5455-11eb-9eca-9dad6fb71ae7.png)

This needs to be customized for matplotlib and Altair.

micahtyong:
Closed via #285.

PR #285, opened 2021-02-22
Improve automatic bin determination for histograms via start, end, and step attributes

### LINKED ISSUE #265: Binned charts should provide start, end, and step
To make sure that Vega-Lite can render histograms correctly, Lux should provide the start and end of a bin as well as the step size. 

Here is an example of a simple bar chart. 

```py
df = pd.DataFrame({'start': [1,2,3,4,5], 'end': [2,3,4,5,6], 'count': [12,52,76,21,32]})

alt.Chart(df).mark_bar().encode(
  x = alt.X('start', bin = alt.Bin(binned=True, step=1)),
  x2 = 'end',
  y = 'count'
)
```

With this core, the space between bars is small (as it should be in a histogram) and the axis ticks on the x axis are spaced like the bars.

Screen Shot 2021-02-15 at 20 07 47

Relevant code https://github.com/lux-org/lux/blob/b12bf54bce412071e171189b7156bbb03a4e302f/lux/vislib/altair/Histogram.py#L66

After this fix, I think you can delete https://github.com/lux-org/lux/blob/b12bf54bce412071e171189b7156bbb03a4e302f/lux/vislib/altair/Histogram.py#L55.

micahtyong:
## In this PR

Closes #265 and #217 by modifying `Histogram.py` and providing a `start`, `end`, and `step` bin for the Altair renderer. The bin width, `step`, is stored in the Vis object during `PandasExecutor.py#execute_binning()`. 

## Changes
-  Modifications in `vislib/altair/Histogram.py` and `executor/PandasExecutor.py`
- Additional instance variable, `.__bin_size`, in `Vis.py`

## Example Output
Here is an updated screenshot for histogram binning along the x-axis. As an example,
```
df['dateBefore'] = pd.to_datetime(df['dateBefore'], format='%Y-%M-%d')
df
```
This yields the following output for Distribution histograms: 
Output

Then, we can specify intent to view Filter histograms. 
```
df.intent = ["severityScore"]
df
```
Screen Shot 2021-03-02 at 9 26 06 PM

Here is another screenshot to show histogram binning along the y-axis, which remains unchanged. 
Screen Shot 2021-02-22 at 1 21 21 AM


lux/vislib/altair/Histogram.py
```diff
@@ -55,38 +55,39 @@ def initialize_chart(self):
...
-                    alt.X(
+                    x=alt.X(
                         str(msr_attr.attribute),
                         title=axis_title,
-                        bin=alt.Bin(binned=True),
+                        bin=alt.Bin(binned=True, step=markbar),
                         type=msr_attr.data_type,
-                        axis=alt.Axis(labelOverlap=True, title=axis_title),
-                        scale=alt.Scale(domain=(x_min, x_max)),
+                        axis=alt.Axis(labelOverlap=False, title=axis_title),
```
domoritz, line 74:
```suggestion
                        axis=alt.Axis(title=axis_title),
```

lux/vislib/altair/Histogram.py
```diff
@@ -55,38 +55,39 @@ def initialize_chart(self):
...
-                .mark_bar(size=markbar)
+                .mark_bar()
                 .encode(
                     x=alt.X("Number of Records", type="quantitative"),
                     y=alt.Y(
                         str(msr_attr.attribute),
                         title=axis_title,
-                        bin=alt.Bin(binned=True),
+                        bin=alt.Bin(binned=True, step=markbar),
                         axis=alt.Axis(labelOverlap=True, title=axis_title),
```
domoritz, line 90:
```suggestion
                        axis=alt.Axis(title=axis_title),
```

lux/vislib/altair/Histogram.py
domoritz: Not sure why the label overlap is set to false here.

lux/vislib/altair/Histogram.py
```diff
@@ -55,38 +55,39 @@ def initialize_chart(self):
         markbar = x_range / plot_range * 12
```
domoritz, line 55: What's the point of this formula? The *12 looks suspicious to me.

micahtyong: From my understanding, `markbar` is Lux's way of determining a consistent bin size across all histograms. Otherwise, Altair will estimate and provide their own bin size. In my recent commit, I change the multiplier to `10`, which yields output like so: 
Screen Shot 2021-02-22 at 12 34 20 PM

When I remove the `markbar` constraint entirely, Altair will interpret the appropriate bin size automatically. This is the example yielded output:

Screen Shot 2021-02-22 at 12 41 56 PM

The key difference is that in the latter input, the bin size is determined for each individual graph. In the former input, the bin size is set in a way such that bin width appears to be the same width across all graphs such that precisely 10 bars are displayed in each graph, giving a more consistent look. From an HCI perspective, I'm not in a position to comment on which one is more meaningful. Perhaps @domoritz you have some input that would help us make a decision on this. 

@dorisjlee, feel free to chime in and correct me if there's anything I misinterpreted.

micahtyong:
Hi @domoritz, thank you for your feedback so far! In my most recent commit (e.g. 9216dba), I explore the latter approach from an earlier discussion where `markbar` is no longer needed. Feel free to comment with your thoughts!

lux/vislib/altair/Histogram.py
domoritz: I think the bins Vega uses are a lot better since they fall on understandable steps but since you pre-aggregate the data, you would need to reimplement the logic in https://github.com/vega/vega/blob/ca27004f8c3ad8aa94f017fb5df6bf90748bda92/packages/vega-statistics/src/bin.js. 

It's probably best to do that later, though, and first merge this fix that sets bin start and end and passes the correct bin size.

12 is fine (as fine as 10). I just think the logic could be more sophisticated to compute bin start and step (using the logic linked above and maybe combined with e.g. Sturge's Rule).

domoritz:
you're very welcome. 

Your latest changes definitely produce better results but require Vega-Lite to do the data transformation. It should work well for small to medium-sized data.

domoritz:
Just to note, the last commit reverts the improvement (for potentially better performance).

dorisjlee:
Thanks @micahtyong! There is some custom bar size determination that we had to do since Lux delegate the binning to the executor side. The new implementation looks good for `Filter` type actions where the xmin/xmax extent is fixed and the filter changes. However, it causes some issues when the extent range is very different, in the case of `Distribution` action. Our earlier implementation worked better in this case.
image

We could look into reimplementing something along the lines of Vega-Lite's [bin.js](https://github.com/vega/vega/blob/ca27004f8c3ad8aa94f017fb5df6bf90748bda92/packages/vega-statistics/src/bin.js.). Thanks for pointing this out @domoritz!

domoritz:
> We could look into reimplementing something along the lines of Vega-Lite's bin.js. Thanks for pointing this out @domoritz!

As a first step, I would suggest passing the step size explicitly to Vega-Lite so that the bars are sized correctly. Then as a next step, you could improve how Lux determines the bin offset and step size.

micahtyong:
# Example Test Suite
The following are examples for reviewers to try. There are also additional tests in `test_vis.py` which validate bin width. To set up, be sure to include
```
import lux
import pandas as pd
lux.config.default_display = "lux"
lux.config.plotting_backend = "vegalite"
```

### Olympic Dataset 
```
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/olympic.csv?raw=True")
df
```
olympic_dist

Now, let's add a filter. 
```
df.intent=["Height"]
df
```
olympic_filter

### Car Dataset
```
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/car.csv?raw=True")
df['Year'] = pd.to_datetime(df['Year'], format='%Y')
df
```
car_dist

```
df.intent=["Acceleration"]
df
```
car_filter

### COVID Dataset
```
df['dateBefore'] = pd.to_datetime(df['dateBefore'], format='%Y-%M-%d')
df
```
covid_dist

```
df.intent = ["severityScore"]
df
```
covid_filter

lux/vislib/altair/Histogram.py
```diff
@@ -45,48 +44,52 @@ def initialize_chart(self):
...
         if isinstance(msr_attr.attribute, str):
             msr_attr.attribute = msr_attr.attribute.replace(".", "")
 
         colval = self.vis.data[msr_attr.attribute]
-        x_range = abs(max(colval) - min(colval))
-        plot_range = abs(x_max - x_min)
-        markbar = x_range / plot_range * 12
+        markbar = get_bin_size(self.data[msr_attr.attribute])
+        # Default when IQR was too small
+        if markbar < (x_range / 24):
```
domoritz, line 55: So when there are fewer than 24 bins, you use 12? Or am I misreading this?

lux/vislib/altair/Histogram.py
```diff
@@ -45,48 +44,52 @@ def initialize_chart(self):
...
+            markbar = (x_max - x_min) / 12
 
         self.data = AltairChart.sanitize_dataframe(self.data)
+        end_attr_abv = str(msr_attr.attribute) + "_end"
+        self.data[end_attr_abv] = self.data[str(msr_attr.attribute)].apply(lambda x: x + markbar)
```
domoritz, line 60: If the data is a data frame, you could just write column + mark around and it would be vectorized (and faster).

lux/vislib/altair/Histogram.py
```diff
@@ -45,48 +44,52 @@ def initialize_chart(self):
...
-                    alt.X(
+                    x=alt.X(
                         str(msr_attr.attribute),
                         title=axis_title,
                         bin=alt.Bin(binned=True),
                         type=msr_attr.data_type,
-                        axis=alt.Axis(labelOverlap=True, title=axis_title),
-                        scale=alt.Scale(domain=(x_min, x_max)),
+                        axis=alt.Axis(title=axis_title),
+                        scale=alt.Scale(domain=[x_min, x_max + markbar]),
```
domoritz, line 76: Why do you need to add the mark bar to the max? Shouldn't the bins be between min and max?

lux/vislib/altair/Histogram.py
```diff
@@ -45,48 +44,52 @@ def initialize_chart(self):
...
                 .encode(
                     x=alt.X("Number of Records", type="quantitative"),
                     y=alt.Y(
                         str(msr_attr.attribute),
                         title=axis_title,
-                        bin=alt.Bin(binned=True),
-                        axis=alt.Axis(labelOverlap=True, title=axis_title),
-                        scale=alt.Scale(domain=(x_min, x_max)),
+                        bin=alt.Bin(binned=True, step=markbar),
+                        axis=alt.Axis(title=axis_title),
```
domoritz, line 92: If you need x2, why not y2 here as well?

lux/vislib/altair/Histogram.py
```diff
@@ -45,48 +44,52 @@ def initialize_chart(self):
 
         x_min = self.vis.min_max[msr_attr.attribute][0]
         x_max = self.vis.min_max[msr_attr.attribute][1]
```
domoritz, line 47: You can write, x_min, x_max = self.vis.min_max[msr_attr.attribute]

domoritz (commented):
The results look good to me. I left a few small comments for improvements.

tests/test_vis.py
```diff
@@ -340,7 +341,9 @@ def test_histogram_chart(global_var):
     lux.config.plotting_backend = "vegalite"
     vis = Vis(["Displacement"], df)
     vis_code = vis.to_Altair()
+    expected_bin_size = get_bin_size(vis.data["Displacement"])
```
dorisjlee, line 344: I still don't exactly understand this. `vis.data` is the data that has already been binned to 10 bins in the PandasExecutor, so in theory, the bin determination would need to go into the Compiler or `execute_binning` stage, if you want access to the raw data series (i.e.`ldf["Displacement"]`). Or is this simply an approximation that you've tested that works well in most cases?

dorisjlee:
Thanks @micahtyong! The results look much better now for histograms in both Distribution and Filter.

lux/vislib/altair/Histogram.py
micahtyong: Here, when the `markbar` is so small that there would be *more than* 24 bins, we default to 12. This occurs when the interquartile range is very small compared to the plot range (e.g. most points fall on one value). Thus, it doesn't happen often when plotting the Distribution tab, but moreso when we plot for the Filter tab. This stems from the fact that we use the `x_min` and `x_max` from the original dataset, not the filtered one, when displaying Filter plots (see the second photo for each of the datasets in the Example Test Suite). For example, in the COVID dataset, adding an intent on severity score would yield a graph like the following. If I didn't add the default, the bar would be very skinny, almost invisible. 

Screen Shot 2021-03-02 at 10 30 16 AM

Let me know if that clarifies things! As always, I'm open to ideas/improvements. As you see, both the `12` and `24` are a bit arbitrary and worked well after testing on multiple datasets. I wonder if there's a more rigorous approach.

lux/vislib/altair/Histogram.py
micahtyong: Thank you for this tip!

lux/vislib/altair/Histogram.py
micahtyong: I ran into an issue where the last bin would exceed the maximum range. For example, if you take a look at the three graphs below after running, 
```
df
```
I get the following output. 
Screen Shot 2021-03-02 at 10 37 09 AM

Thus, I added markbar to account for the bin width. ?

lux/vislib/altair/Histogram.py
micahtyong: Great tip, thanks!

micahtyong: Great tip, thanks!

lux/vislib/altair/Histogram.py
domoritz: Thanks for the explanation. I totally got the comparison inverted in my head.

tests/test_vis.py
micahtyong: I took a look at `execute_binning` and I think I have an explanation (albeit an amateurish one). `execute_binning` successfully divides up the data into 10 bins using `np.histogram` and computes the centers where each of these bins would lie. However, what `execute_binning` fails to do is compute the bin widths. Thus, `get_bin_size` recomputes the bin width according to Freedman's Rule. 

But actually, this is a great point, I realized that `execute_binning` computes a secondary variable called `bin_edges`, which I could use to extract the bin width and store in the Vis object as `._bin_width`. In my most recent commit (#5b4643c), I experiment with this approach, and it yields the same results! Let me know if this is a preferred method, and I will update `test_vis.py` accordingly. @dorisjlee

lux/vislib/altair/Histogram.py
domoritz: Something looks wrong in those charts. The ticks are not between the bars as they should be. 

Can you send me one of those specs as a Vega-Lite spec? I can give you better feedback that way.

micahtyong: I think I've figured it for Distribution charts following from my realization [here](https://github.com/lux-org/lux/pull/285#discussion_r585924162). Specifically, ticks are now between the bars. Does this look better? 
Screen Shot 2021-03-02 at 1 59 25 PM

Not super sure what a "Vega-Lite spec" is, but here is the output from running `to_Altair()` on one of the visualizations: 
```
chart = alt.Chart(visData).mark_bar(size=3.8899999999999997).encode(
    alt.X('severityScore', title='severityScore (binned)',bin=alt.Bin(binned=True), type='quantitative', axis=alt.Axis(labelOverlap=True, title='severityScore (binned)'), scale=alt.Scale(domain=(0.1, 39.0))),
    alt.Y("Number of Records", type="quantitative")
)
```

lux/vislib/altair/Histogram.py
```diff
@@ -69,10 +69,10 @@ def initialize_chart(self):
                     x=alt.X(
                         str(msr_attr.attribute),
                         title=axis_title,
-                        bin=alt.Bin(binned=True),
+                        bin=alt.Bin(binned=True, step=markbar),
```
domoritz, line 72: 👍

tests/test_vis.py
dorisjlee: Thanks for the explanation @micahtyong! This makes a lot of sense. I was somewhat misled into thinking that `get_bin_size` computes the number of bins, instead of the bin width, which explains my confusion in thinking that this should go into the Executor. The original approach that you took seems like the more sensible approach. I'm glad that you were able to try this via the numpy `bin_edges` approach and verify that the results were similar. For the purpose of integrating with other types of executor (e.g. SQL), the bin width determination makes sense with the Freedman's Rule helper function that you had earlier.

dorisjlee (approved):

domoritz (approved):

dorisjlee:
Thanks @micahtyong, the new histograms look great! I think the cutoff that you included for the bin width really helped with the low number of datapoints filter examples. 
Special thanks @domoritz for all the helpful feedback!

domoritz:
Thank you @micahtyong for the improvements to binning.

micahtyong:
Thank you @domoritz for all the helpful feedback, both technical and from your HCI expertise! I really enjoyed the initiated conversations from this PR. 

Thank you @dorisjlee for the feedback, review, and pointer to the executor! I learned a lot from this enhancement.

PR #286, opened 2021-02-24
Extend color legend size for scatter charts for matplotlib

caitlynachen:
Changed figsize for scatter charts
Screen Shot 2021-02-23 at 5 53 15 PM
Screen Shot 2021-03-07 at 9 04 17 PM


dorisjlee:
Thanks @caitlynachen! This looks like a good start. Here are a couple of cases where the figure size doesn't quite work out. The plot figure looks overly stretched out. We might also need to abbreviate the legend when the number of colors is too large.
image
image

lux/vislib/matplotlib/ScatterChart.py
```diff
@@ -57,14 +57,14 @@ def initialize_chart(self):
...
         if len(color_attr) == 1:
-            self.fig, self.ax = matplotlib_setup(6, 5)
             color_attr_name = color_attr[0].attribute
             color_attr_type = color_attr[0].data_type
             colors = df[color_attr_name].values
             plot_code += f"colors = df['{color_attr_name}'].values\n"
             unique = list(set(colors))
             vals = [unique.index(i) for i in colors]
             if color_attr_type == "quantitative":
+                self.fig, self.ax = matplotlib_setup(7, 5)
```
dorisjlee, line 67: The code inside matplotlib_setup doesn't show up when we export the chart. As a result, the figure size in the exported code and the widget is different. Is this the intended behavior?

caitlynachen: fixed

caitlynachen:
fixed corner cases @dorisjlee mentioned

Screen Shot 2021-02-24 at 10 40 11 AM
Screen Shot 2021-02-24 at 10 39 52 AM

dorisjlee:
Thanks @caitlynachen! A few more comments on the latest changes: 

- For the abbreviated chart, could we do "..." to indicate that there is more color in the legend than we are displaying? In the `Brand` legend example, there are only 16 colors displayed in the legend but there is 29 different `Brand` in the dataset.

- When we change the width and height of the resulting plot, the fontsize becomes much smaller. Is there a way we can ensure that the font and graphics are consistent across the charts (regardless of size of chart)?
image

lux/vislib/matplotlib/ScatterChart.py
```diff
@@ -82,10 +84,29 @@ def initialize_chart(self):
...
-                scatter = self.ax.scatter(x_pts, y_pts, c=vals, cmap="Set1")
-                plot_code += f"scatter = ax.scatter(x_pts, y_pts, c={vals}, cmap='Set1')\n"
+                if len(unique) >= 16:
+                    unique = unique[:16]
+
+                maxlen = 0
+                for i in range(len(unique)):
+                    unique[i] = str(unique[i])
+                    if len(unique[i]) > 26:
+                        unique[i] = unique[i][:26] + "..."
```
dorisjlee, line 94: This is not showing up.

caitlynachen: what do you mean by this?

dorisjlee: I meant the abbreviation "..." is not showing up in the legend, see comment above.

caitlynachen: o hmm i think it was working on my end
109049161-c787d900-768c-11eb-95d0-44363d65fabc

caitlynachen: are you seeing something differently

jerrysong1324: I'm able to see it.

jerrysong1324 (approved):

lux/vislib/matplotlib/ScatterChart.py
dorisjlee: image

Oops sorry, I meant when the `...` vertically (not horizontally) when there are too many color categories.

>  In the Brand legend example, there are only 16 colors displayed in the legend but there is 29 different Brand in the dataset.

jerrysong1324: Oh this is a good point, a user might be confused why a specific category is not showing up, so something in the legend would be nice to address this. Another idea rather than using the '...' (which seems a bit off to me) could be to use multiple columns. Try maybe something like this: https://stackoverflow.com/questions/58344791/how-to-handle-categorical-legends-with-too-many-entries-to-match-a-fixed-chart-h

caitlynachen: ohh the "..." is not for the length of the legend, but rather the individual categories. I checked and Altair also didn't have a "..." at the end of the list, so didn't include it in this PR. If you want, maybe I can tackle that issue separately

caitlynachen: Added an issue https://github.com/lux-org/lux/issues/295 for this new change

dorisjlee:
Hi @caitlynachen, Thanks for adding in #295. Could you look into addressing the earlier comment on consistency across fontsizes when the image stretching happens?

> * When we change the width and height of the resulting plot, the fontsize becomes much smaller. Is there a way we can ensure that the font and graphics are consistent across the charts (regardless of size of chart)?
>   image

caitlynachen:
Added fontsize to xlabel, ylabel and legend attributes to maintain fontsize to be as large as possible within the range of plt.tight_layout

caitlynachen:
In response to @dorisjlee 's comment about reduced font size on these scatter charts with color legends, I tried a number of solutions:

1. Attempting to increase the figure size (height and width)
2. Setting rcParams after creating larger figure size to larger font size
3. Removing plt.tight_layout() to remove automatic sizing
4. Setting plt.gcf().subplots_adjust(bottom=0.15) and plt.gcf().subplots_adjust(left=0.15) to add additional padding space for labels
5. Setting plt.rcParams['figure.autolayout'] = True to fit charts to figure size

Despite the changes, the font size still remained smaller than normal. 

Screen Shot 2021-03-08 at 9 49 05 AM

Not entirely sure why this is the case, especially since when I print out `plt.rcParams['axes.labelsize']`right before plt.savefig, it says the font is as expected at 16, but doesn't display as such. Let me know if anyone else is familiar with Matplotlib frontend and would like to help out on this issue!

Resources:
https://stackoverflow.com/questions/6774086/why-is-my-xlabel-cut-off-in-my-matplotlib-plot
https://matplotlib.org/3.1.0/api/matplotlib_configuration_api.html

dorisjlee (approved):

dorisjlee:
Thanks for the explanation @caitlynachen! I'm merging this in for now, for the matplotlib font size issue, could you create a separate issue outlining this problem? I believe this is a more general issue with the font size and figures for matplotlib and is not particular to just the axis label (affects all text and elements on the figure when the figure is resized).

PR #291, opened 2021-03-02
Lazy Streaming of Tabs

cjachekang:
## Overview

This change (along with the frontend changes) implements lazy loading of the tabs. In essence, one tab is calculated at a time and displayed at a time.

## Changes

I had to change a lot of the tests in terms of the tabs they were expecting since we are now lazily streaming them in, the order could be changed (I just changed the tests from using lists to using sets).  I changed the way we compute actions. Instead of doing them all at once, I compute the first available one and display that (so we have anything to display). Then when repr_html reaches its end, I start calculating the rest of the tabs.

## Example Output

Please see frontend PR.


jerrysong1324 (approved):

dorisjlee:
Hi @cjachekang, Thanks for adding these in, the streaming implementation looks really great! The interactivity that we now have with this is incredible! Here are some notes and comments: 

1. Can you add in some description in the PR on the performance benefits of these changes? 
2. One unexpected thing that I saw is that the warnings are now displayed under the widget, as opposed to above like before. This is not necessarily a bad thing, I'm just curious why this is happening. 
image
3. Also I'm wondering if there is a way to better style the tabs. I'm personally not a huge fan of the Pandas/Lux tabs nested in the action tabs, especially given that the two types of tabs look very different (e.g., one in ipywidget and one in JS). In added the nested Lux widget takes up a lot of unnecessary space with the double border one the left and right side. Is there something that we can do in the CSS that make this more closely resemble the toggle button that we had earlier (e.g. invisible div + buttons)?
4. For the tabs that are streaming in, I'm wondering if showing some type of indicator that additional tabs are still being computed would be helpful. I remember we briefly experimented with the idea of the greyed-out tab that becomes active again after the results are in, is there a reason why we moved away from this implementation?
5. Temporal shows up first which usually just involves a single visualization. I'm wondering if there is a way to show something that contains a couple more visualizations in the initial view.

cjachekang:
@dorisjlee 

2. The warnings are displayed below the widget now since I initially display the pandas dataframe/tabs widget before any computations start. 

3. I'll play around with different stylings and see what look's good.

4. The reason is that it was becoming hard to know what tabs (at least temporal) were actually going to be computed before we computed it. I saw Will's demo in the research meeting and he had a bunch of new tabs. I think for extensibility purposes, it'll be easier (and will save dev time) to push out tabs once they're done since we won't have to write code to preprocess the each new action (to know if it is going to be displayed) everytime we add one.

5.. Yeah! The tabs can be computed in any order - for now I just reversed the order so that correlation is the last to be computed. How do you want me to prioritize the actions? They come in alphabetical as default.

jerrysong1324:
I personally think the old tab method was a bit better. Maybe you can just change the font to match what is within our widget as well as change the paradigm for when the table is not selected. (Our widget uses blue text, the ipython tabs seem to use grey backgrounds).

lux/core/frame.py
```diff
@@ -90,6 +90,7 @@ def __init__(self, *args, **kw):
         self._min_max = None
         self.pre_aggregated = None
         self._type_override = {}
+        self.loadingBar = None
```
dorisjlee, line 93: Let's rename this as `loading_bar` to be consistent with PEP8.

lux/core/frame.py
```diff
@@ -426,30 +428,55 @@ def maintain_recs(self, is_series="DataFrame"):
...
-                if len(vlist) > 0:
-                    rec_df._recommendation[action_type] = vlist
+                rec_df._recommendation[action_type] = vlist
             rec_df._rec_info = rec_infolist
             rec_df.show_all_column_vis()
-            if lux.config.render_widget:
-                self._widget = rec_df.render_widget()
+
+            self._widget = rec_df.render_widget(
+                pandasHtml=rec_df.to_html(max_rows=5, classes="pandasStyle")
```
dorisjlee, line 469: Let's make this 10 so that it corresponds to the default [`display.min_rows` value](https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html) which is 10

lux/core/frame.py
```diff
@@ -426,30 +428,55 @@ def maintain_recs(self, is_series="DataFrame"):
...
+            )
+
         # re-render widget for the current dataframe if previous rec is not recomputed
         elif show_prev:
             rec_df.show_all_column_vis()
             if lux.config.render_widget:
-                self._widget = rec_df.render_widget()
+                self._widget = rec_df.render_widget(
+                    pandasHtml=rec_df.to_html(max_rows=5, classes="pandasStyle")
```
dorisjlee, line 477: same as above

lux/core/frame.py
```diff
@@ -675,11 +722,15 @@ def render_widget(self, renderer: str = "altair", input_current_vis=""):
         import luxwidget
 
         widgetJSON = self.to_JSON(self._rec_info, input_current_vis=input_current_vis)
+        if pandasHtml is None:
```
dorisjlee, line 725: This is for the purpose of backwards compatibility with other versions of lux-api right?

lux/action/custom.py
```diff
@@ -45,36 +45,67 @@ def custom(ldf):
...
-        return []
+                if lux.config.actions[action_name].args:
+                    if not lux.config.actions[action_name].args[0] in data_types:
+                        continue
+                keys.append(action_name)
+            progress += 1
+            if loadingBar is not None:
+                loadingBar.value = progress
+
+    # # Pushing back correlation and geographical actions for performance reasons
```
dorisjlee, line 102: How do we currently determine what to compute first v.s. what to compute lazily? Or for now, are we explicitly saying that "correlation" and "geographical" will compute later?

dorisjlee (changes requested):

lux/core/frame.py
cjachekang: Perhaps - there was an error in one of the tests that caused pandasHtml to be none, so I just put a null check here

lux/action/custom.py
cjachekang: Right now, actions are checked in alphabetical order (with correlation being first). I manually put correlation and geographical later because i noticed those two take the longest to compute/render. The order in which we check/compute is very flexible as we can just move the action names around in the list.

lux/core/frame.py
dorisjlee: Yeah, when we change the input to luxwidget, the missing traitlet variable can cause an error. So It's good that we have this check in here.

PR #294, opened 2021-03-03
Adding python-app.yml for Github Action CI

jinimukh:
## Overview

This PR includes a GitHub Actions File that will take the place of the `.travis.yml` file.

## Changes
- Add `.github/workflows/python-app.yml`
- Remove `.travis.yml`


jinimukh:
After we merge this in, we should delete the branch! @dorisjlee

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,40 @@
...
+        pip install -r requirements-dev.txt
+    - name: Lint check with black
+      run: |
+        black --target-version py37 --line-length 105 --check .
+    - name: Test with pytest
+      run: |
+        python -m pytest tests/*.py
+    - name: Code Coverage Report
+      run: |
+        pytest --cov-report term --cov=lux tests/
```
dorisjlee, line 39: I know that this is probably copied over from what we originally had in the `.travis.yml`, but it looks like this line rerun the test suite again with the coverage report? 
Could we combine this line with the earlier `python -m pytest tests/*.py` so that we only need to run the test suite once?

jinimukh: Let me check.

jinimukh:
@dorisjlee This is ready to merge -- failing pytest causes the exit code returned to be 1!

PR #301, opened 2021-03-12
Revert Cardinality Requirement for Histograms

jinimukh:
## Overview

Now that we have a new mechanism of finding bin widths, we can remove the cardinality requirement that was put into place to fix a divide-by-zero error that occurred if the range of values was 0.

## Changes
- remove cardinality requirement in `univariate.py`
- take the absolute value for the default markbar in `histogram.py`
- adjust tests to current output

## Example Output

Now, we can see uniform distributions. Here is an example:

Screen Shot 2021-03-11 at 11 33 23 PM


lux/vislib/altair/Histogram.py
```diff
@@ -53,7 +53,7 @@ def initialize_chart(self):
 
         # Default when bin too small
         if markbar < (x_range / 24):
-            markbar = (x_max - x_min) / 12
+            markbar = abs(x_max - x_min) / 12
```
micahtyong, line 56: Would there ever be a case such that calling `abs` is necessary? By construction, shouldn't `x_max` be greater than or equal to `x_min`?

lux/action/univariate.py
```diff
@@ -46,9 +46,7 @@ def univariate(ldf, *args):
     ignore_rec_flag = False
     if data_type_constraint == "quantitative":
         possible_attributes = [
-            c
-            for c in ldf.columns
-            if ldf.data_type[c] == "quantitative" and ldf.cardinality[c] > 5 and c != "Number of Records"
+            c for c in ldf.columns if ldf.data_type[c] == "quantitative" and c != "Number of Records"
```
micahtyong, line 49: Just to clarify, does `ldf.cardinality[c]` return the number of rows for a given column? In this case, I wonder if removing this check would lead to some irregular behavior when there are very few rows. Is there some other part in the code that checks for this?

lux/vislib/altair/Histogram.py
jinimukh: I agree, I added this as a final check for integer overflow so that at least our code doesn't error and it just means the user inputs are too extreme (positive or negative) to handle. @dorisjlee should we even worry about this?

lux/action/univariate.py
jinimukh: no. it returns the unique number of values -- we only compute histograms for dataframes > 5 rows.

dorisjlee:
Thanks @jinimukh! Could you explain in the PR what is the benefits of removing the cardinality check? In particular, what was the issue the original implementation did for columns with cardinality lower than 5? Would the charts simply not display or is there an error?  Also this change seems like it was applied to both bars and histograms (for both quantitative and nominal in univariate.py)? If that's the case, we can update the description accordingly.

lux/vislib/altair/Histogram.py
dorisjlee: I think xmax will always be larger than xmin.

jinimukh:
@dorisjlee If we had cardinality lower than 5, the histogram/bar graph would not show. We talked about this with respect to series — if they all have the same value, I think it would be nice to show a uniform distribution. Was there an initial reason to not even compute it in the first place other than having one (or a few) less visualizations to compute?

dorisjlee:
Yes I think in this case showing the uniform distribution (as a bar chart) would make sense. I think the original concern was that if it was a single column with floats, i.e. Units = 4.0, it might be recognized as a quantitative column --> plotting a histogram where range is 0 causes errors (I believe we already fixed this in #99). Could you add a few low cardinality test cases case for each of the data types (quantitative, nominal, etc.) that you modified part of the PR?

dorisjlee:
Thanks @jinimukh!

PR #306, opened 2021-03-14
Adding support for SQL as execution engine

thyneb19:
## Overview

Merging in the sql-engine branch to bring the updated Lux SQL functionality to the main release. Users will be able to connect LuxSQLTable objects to their database tables and views, and leverage Lux' recommendation system without having to pull all of their database data locally.

## Changes

This PR will update the Lux SQLExecutor as well as add a new LuxSQLTable object. The LuxSQLTable object is meant to help users differentiate between Lux' functionality when connecting to a Dataframe versus a SQL database. This PR makes the following major changes:

- Updates SQLExecutor.py to query data necessary for all of Lux' supported charts
- Adds sqltable.py which contains the LuxSQLTable class. This class inherits the Lux recommendation system utilities from the LuxDataFrame object
- Adjusts frame.py to ensure metadata and recommendation maintenance works when using the SQLExecutor
- Adds scripts to test the functionality of the SQLExecutor

## Example Output

Script to reproduce: 
After setting up Postgres, go into command line via `psql postgres`, then setup via:
```sql
CREATE USER postgres WITH PASSWORD 'lux';
ALTER USER postgres WITH SUPERUSER;
DROP schema public cascade;
CREATE schema public;
CREATE DATABASE postgres;
```
Then run `python upload_car_data.py` or other data upload scripts inside `lux/data`

```python
from sqlalchemy import create_engine
engine = create_engine("postgresql://postgres:lux@localhost:5432")

tbl = lux.LuxSQLTable()
lux.config.set_SQL_connection(engine)
tbl.set_SQL_table("car")

tbl
```
Users will now be able to connect Lux to their database tables like so:

sql_executor


dorisjlee:
Thanks @thyneb19! I'll do a closer review in a bit, but several quick items to note. Could we move the ipynb files to a different repo and add a link to this? We might also have to modify the Github action script for the test similar to what you did with the postgres setup in the old .travis.yml.

lux/_config/config.py
```diff
@@ -358,9 +359,10 @@ def set_executor_type(self, exe):
             from lux.executor.SQLExecutor import SQLExecutor
 
             self.executor = SQLExecutor()
-        else:
+        elif exe == "Pandas":
             from lux.executor.PandasExecutor import PandasExecutor
 
+            self.SQLconnection = ""
             self.executor = PandasExecutor()
```
jinimukh, line 367: maybe worthwhile to throw an Error here if the execute is neither Pandas nor SQL.

lux/core/frame.py
```diff
@@ -168,12 +179,15 @@ def _infer_structure(self):
...
+        small_df_flag = len(self) < 100 and lux.config.executor.name == "PandasExecutor"
         if self.pre_aggregated == None:
             self.pre_aggregated = (is_multi_index_flag or not_int_index_flag) and small_df_flag
             if "Number of Records" in self.columns:
                 self.pre_aggregated = True
-            self.pre_aggregated = "groupby" in [event.name for event in self.history]
+            self.pre_aggregated = (
+                "groupby" in [event.name for event in self.history]
+                and lux.config.executor.name == "PandasExecutor"
+            )
```
jinimukh, line 191: why do the checks have to be for `lux.config.executor.name`? Can these attributes not be computed by the SQLTable? If so, should we flip the order of the statements so it short-circuits?

lux/action/univariate.py
```diff
@@ -104,7 +104,7 @@ def univariate(ldf, *args):
             "long_description": "Temporal displays line charts for all attributes related to datetimes in the dataframe.",
         }
         # Doesn't make sense to generate a line chart if there is less than 3 datapoints (pre-aggregated)
-        if len(ldf) < 3:
+        if ldf.length < 3:
             ignore_rec_flag = True
```
jinimukh, line 108: I'm a little confused for the change from `len(ldf)` to `ldf.length`. Could you please explain? I don't think Pandas DataFrame has the attribute `length` so I'm worried that this will error, and confused why it won't :P

lux/core/sqltable.py
```diff
@@ -0,0 +1,192 @@
...
+        self.unique_values = None
+        self.cardinality = None
+        self._min_max = None
+        self.pre_aggregated = None
+        self._type_override = {}
+
+        if table_name != "":
+            self.set_SQL_table(table_name)
+        warnings.formatwarning = lux.warning_format
+
```
jinimukh, line 89: Why do we have a separate `init` function. It looks like `frame.py` takes care of both `SQL` case and `Pandas` case, so why can't we call `super()` and then change whatever we need to?

lux/core/sqltable.py
```diff
@@ -0,0 +1,192 @@
...
+        except Exception as error:
+            error_str = str(error)
+            if f'relation "{t_name}" does not exist' in error_str:
+                warnings.warn(
+                    f"\nThe table '{t_name}' does not exist in your database./",
+                    stacklevel=2,
+                )
+
+    def _repr_html_(self):
+        from IPython.display import display
```
jinimukh, line 112: I assume this will have to change with @westernguy2 's changes. Is there some way you can again re-use `super._repr_html_()` or what it becomes as per #308 ?

lux/executor/PandasExecutor.py
```diff
@@ -515,6 +515,7 @@ def compute_stats(self, ldf: LuxDataFrame):
         ldf.unique_values = {}
         ldf._min_max = {}
         ldf.cardinality = {}
+        ldf.length = len(ldf)
```
jinimukh, line 519: Ahh I understand that `ldf.length` now exists, but it doesn't always exist. Can we set this upon initialization?

tests/test_config.py
```diff
@@ -51,6 +52,7 @@ def contain_horsepower(df):
 
 def test_default_actions_registered(global_var):
     df = pytest.car_df
+    lux.config.set_executor_type("Pandas")
     df._repr_html_()
```
jinimukh, line 56: why do we need line 55? If it is set to `SQL` those methods should handle the issue? Maybe the expected will change, but I'm not sure.

tests/test_interestingness.py
```diff
@@ -298,6 +298,7 @@ def test_interestingness_deviation_nan():
 
     smaller_diff_score = interestingness(vis, test)
     bigger_diff_score = interestingness(vis2, test)
-    assert np.isclose(smaller_diff_score, 0.29, rtol=0.1)
-    assert np.isclose(bigger_diff_score, 0.94, rtol=0.1)
+    print(smaller_diff_score, bigger_diff_score)
+    assert np.isclose(smaller_diff_score, 0.19, rtol=0.1)
```
jinimukh, line 302: I dont think we need the print statement.

tests/test_vis.py
```diff
@@ -458,6 +458,7 @@ def test_vegalite_default_actions_registered_2(global_var):
 
 
 def test_matplotlib_default_actions_registered(global_var):
+    df = pytest.car_df
     lux.config.plotting_backend = "matplotlib"
```
jinimukh, line 462: why do we need 461 and 463? what changed?

jinimukh (commented):
Absolutely amazing! So far, I was only able to look through the code -- I'll play around with it sometime tonight/tomorrow morning. Great work though! This is super cool. One small note: do we need to merge in `sql_benchmarking.csv` or can we get rid of that?

tests/test_interestingness.py
dorisjlee: Yup I agree, we can take this out.

tests/test_vis.py
thyneb19: Yeah that is redundant, will remove one of the unnecessary calls here.

tests/test_interestingness.py
thyneb19: Yes looks like something that was accidentally left in during some bug testing, will remove.

tests/test_config.py
thyneb19: In the test_compiler script we set the global executor type to be SQLExecutor in the last test so we need to change the executor type back to Pandas here, otherwise none of the data gathering will work with the dataframe and a warning will be thrown.

Looking back at test_config, the `set_SQL_connection()` call in `register_new_action()` acctually seems redundant. For this script we just need to set the executor to be a PandasExecutor in the first test function.

NiStannum:
@dorisjlee Nitin here

CONTRIBUTING.md
```diff
@@ -57,6 +57,9 @@ python -m pytest tests/<test_file_name>.py
 
 
 
+# Code Formatting
+In order to keep our codebase clean and readible, we are using PEP8 guidelines. To help us maintain and check code style, we are using [black](https://github.com/psf/black). Simply run `black --line-length 105 .` before commiting. Failure to do so may fail the tests run on Travis. This package should have been installed for you.
```
westernguy2, line 61: Nit:  but we aren't using Travis anymore right? Also I think it's just `black .`?

lux/core/series.py
```diff
@@ -43,6 +43,7 @@ class LuxSeries(pd.Series):
         "_prev",
         "_history",
         "_saved_export",
+        "name",
```
westernguy2, line 46: "name" is already added to the bottom of this list (will be added twice)

lux/core/sqltable.py
```diff
@@ -0,0 +1,192 @@
...
+
+            button = widgets.Button(
+                description="Toggle Data Preview/Lux",
+                layout=widgets.Layout(width="200px", top="5px"),
+            )
+            self.output = widgets.Output()
+            lux.config.executor.execute_preview(self)
+            display(button, self.output)
+
+            def on_button_clicked(b):
```
westernguy2, line 148: Will the widget and display have to be re-done whenever things like `_repr_html_` change in general? For example, #291 might change the `_repr_html_` quite a bit.

lux/executor/PandasExecutor.py
```diff
@@ -501,7 +501,7 @@ def _is_geographical_attribute(series):
 
     @staticmethod
     def _is_datetime_number(series):
-        if series.dtype == int:
+        if series.dtype == int or "int" in str(series.dtype):
```
westernguy2, line 504: Why is the second condition needed? And if it is needed, shouldn't there be a cleaner way to search if the Series has an `int` dtype?

tests/conftest.py
```diff
@@ -23,6 +23,7 @@ def global_var():
         "_prev",
         "_history",
         "_saved_export",
+        "name",
```
westernguy2, line 26: Same thing as before, the "name" is added already

tests/test_series.py
```diff
@@ -40,6 +40,7 @@ def test_df_to_series():
         "_prev",
         "_history",
         "_saved_export",
+        "name",
```
westernguy2, line 43: Same as before, "name" already added!

westernguy2 (changes requested):
Looks really great Thyne! It'll be very exciting when this is merged in!

dorisjlee:
Hi @NiStannum, Do you want to take a stab at updating the [ReadTheDocs documentation page for the SQL feature](https://lux-api.readthedocs.io/en/latest/source/advanced/executor.html)? You would need to modify the description inside [executor.rst](https://github.com/lux-org/lux/blob/master/doc/source/advanced/executor.rst).

CONTRIBUTING.md
dorisjlee: I fixed this in the CONTRBUTING doc.

requirements.txt
```diff
@@ -5,8 +5,8 @@ pandas>=1.2.0
 scikit-learn>=0.22
 matplotlib>=3.0.0
 # Install only to use SQLExecutor
-# psycopg2>=2.8.5
-# psycopg2-binary>=2.8.5
+psycopg2>=2.8.5
```
dorisjlee, line 8: Does our code depend on these or are they user-specified? In the latter case, we can make users install these dependencies on their own. Since users have a choice of using `psycopg2` or `sqlalchemy`, we can simply let the user install these connections on their own.

lux/vis/Vis.py
```diff
@@ -351,6 +351,17 @@ def refresh_source(self, ldf):  # -> Vis:
             self._source = ldf
             self._inferred_intent = Parser.parse(self._intent)
             Validator.validate_intent(self._inferred_intent, ldf)
+            vlist = [Compiler.compile_vis(ldf, self)]
```
dorisjlee, line 354: This seems redundant with the `compile_vis` and `execute` that happens subsequently. Is there a reason why we are doing this twice?

lux/vislib/altair/AltairRenderer.py
```diff
@@ -51,10 +51,9 @@ def create_vis(self, vis, standalone=True):
         """
         # Lazy Evaluation for 2D Binning
         if vis.mark == "scatter" and vis._postbin:
-            vis._mark = "heatmap"
-            from lux.executor.PandasExecutor import PandasExecutor
-
-            PandasExecutor.execute_2D_binning(vis)
+            if lux.config.executor.name == "PandasExecutor":
```
dorisjlee, line 54: It looks like the SQLExecutor supports 2D binning for heatmaps, why do we only do this for PandasExecutor?

tests/test_interestingness.py
dorisjlee: I removed this from the test file.

dorisjlee:
@NiStannum: I had a question about the caption. In one of the screenshots that you sent, there was a caption noting that the data table was a preview. I had some initial concerns around how the caption was showing for both the data preview table and the Lux view (and the caption text doesn't seem very applicable in the Lux view case), but it seems like the caption is not part of this PR?
image

@thyneb19: We discussed how the export feature does not completely work currently, since the dataframe is empty. Before we add in the SQL query export capabilities later on, one quick workaround is to default to adding `standalone=True` option for all exports when the SQLExecutor is used. That way the visualization code itself would contain the raw data, instead of having to depend on df (which is currently empty).
```python
print(tbl.exported[0].to_code(language="altair",standalone=True))
```

image

sophiahhuang:
@dorisjlee hello

lux/core/sqltable.py
thyneb19: Oh yeah looks like the `init` function in `frame.py` still has the set `table_name` initialization in it. We can actually remove that now that we have a separate LuxSQLTable object to handle the database use case.

Other than that you're right, we can totally just call `super()` and just have the `table_name` set up. Will change this

dj-khandelwal:
Commenting on the PR! :)

lux/executor/PandasExecutor.py
thyneb19: @dorisjlee So currently the `ldf.length` parameter is being updated whenever `compute_stats` is called by maintain metadata. Whenever a dataframe is manipulated in some way, this parameter is adjusted alongside all of the other parameters like `_min_max`, `unique_values`, etc.

dorisjlee: It should be, but we would need to make sure that it doesn't break anything. Ideally, we should [override the `__len__` variable](https://stackoverflow.com/questions/30483752/is-it-possible-to-modify-the-behavior-of-len) inside the `LuxSQLTable` to change what we get from len(tbl). That way we would not have to change the behavior of `PandasExecutor` or `LuxDataFrame`, which has more potential for causing errors down the line.

thyneb19: Oh yeah overriding `__len__` makes a lot of sense since we could keep the same syntax for the `LuxSQLTable` and `LuxDataFrame` objects. I'll look into trying that out

lux/core/sqltable.py
thyneb19: Yeah looking at #291 we would also probably have to change `_repr_html_` here to work with lazy streaming of tabs. In general most of the changes in `frame.py` from #291 can probably carry over, but we would need to adjust the default Pandas view.

lux/core/series.py
thyneb19: Oh cool, good to know. Will remove this then

lux/executor/PandasExecutor.py
thyneb19: Yeah so there seems to be some different "int" labels when pulling data from SQL databases. When doing some testing we were getting numpy datatypes like `int64` which was failing this check

requirements.txt
dorisjlee: Merging this in for now, but we should check that `psycopg2` doesn't break for installing on different systems.

tests/test_compiler.py
```diff
@@ -17,6 +17,7 @@
 import pandas as pd
 from lux.vis.Vis import Vis
 from lux.vis.VisList import VisList
+import psycopg2
```
dorisjlee, line 20: Can we make the SQL tests into a different file so that the separation between SQL and Pandas tests is more clear? That way, it would make it easier to add a flag that only runs PandasExecutor tests (e.g., when developing locally).

dorisjlee: On a related note, the line `connection = psycopg2.connect("host=localhost dbname=postgres user=postgres password=lux")` is repeated many times across the tests. We could probably put this inside `conftest.py` as a `@pytest.fixture` (variables that are reused across tests in the same session).

tests/test_type.py
```diff
@@ -21,6 +21,7 @@
 
 # Suite of test that checks if data_type inferred correctly by Lux
 def test_check_cars():
+    lux.config.set_SQL_connection("")
```
dorisjlee, line 24: Why is this empty?

dorisjlee (changes requested):

dorisjlee:
Thanks @thyneb19 @dj-khandelwal @NiStannum @sophiahhuang  for working hard on the SQL features! We're really looking forward to this new addition to the upcoming release!

Issue #60, opened 2020-08-11
Adding tests to ensure Pandas function coverage


dorisjlee:
Add test suite to ensure that metadata and recommendations maintained for a variety of Pandas functions.

Issue #56, opened 2020-08-06
Better __repr__ for LuxDataFrame


dorisjlee:
`__repr__` is currently empty to facilitate better display(..) of the luxWidget. We should support string `__repr__` of LuxDataFrame for string operations on the LuxDataFrame. 
One toy example of where this shows up is when dataframes are put into a list, the string `__repr__` is triggered:
![image](https://user-images.githubusercontent.com/5554675/89502597-d6de2c80-d7f7-11ea-96c8-39d360de1f7f.png)

Ideally, we should have a string version that looks more like this:
![image](https://user-images.githubusercontent.com/5554675/89502875-448a5880-d7f8-11ea-9e58-4a7c4cfabbb2.png)

dorisjlee:
Closed with #308

PR #308, opened 2021-03-15
Move widget display to `Out[...]`

westernguy2:
## Overview

The widget display used to be displayed before Out[...], which is not in accordance to what Pandas usually does. This PR moves the widget to Out[...] in IPython notebooks. This PR addresses both #56 and #60

## Changes

I replaced `_repr_html_` with `_ipython_display_` in both `frame.py` and `vis.py`. I then removed `__repr__` from `frame.py` so that it uses the Pandas default `__repr__` when the widget isn't displayed. I added a test to ensure this is the case.

## Example Output
Below is an example of the widget showing up in Out[...]. Notice that Out[...] doesn't show up below the widget, which means the widget is actually in Out[...]

Screen Shot 2021-03-14 at 5 02 09 PM

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -317,10 +317,6 @@ class LuxDataFrame(pd.DataFrame):
     def current_vis(self, current_vis: Dict):
         self._current_vis = current_vis
 
-    def __repr__(self):
-        # TODO: _repr_ gets called from _repr_html, need to get rid of this call
-        return ""
-
     #######################################################
     ########## SQL Metadata, type, model schema ###########
     #######################################################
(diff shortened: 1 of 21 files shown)


cjachekang (approved):

dorisjlee:
Hi @westernguy2, The improved output looks much more Pandas-consistent than what we had before! Could we use the 
`_ipython_display_` approach also for series.py? That way the print(series) behavior would be closer to what print(df) does.

@cjachekang, @westernguy2: Were you able to test this on Jupyter Lab? I tried this on lab but the widget doesn't seem to be displaying. It is possible that there is something wrong with my lab setup.

dorisjlee (approved):

dorisjlee:
This looks great, thanks @westernguy2!

PR #309, opened 2021-03-15
Refactor LuxGroupBy using multiple inheritance

westernguy2:
## Overview

Refactored `LuxGroupBy` to look similar to how Pandas structures GroupBy objects, separating them in `LuxDataFrameGroupBy` and `LuxSeriesGroupBy`. This will allow us to extend methods for specific types of GroupBy objects.

Also fixed a minor bug for testing purposes.

## Changes

Used multiple inheritance to ensure each `LuxGroupBy` child class also inherits the corresponding Pandas version. They all have a parent class of `LuxGroupBy`. Also fixed a bug where the Pandas tests were not using Lux since it was using a different type of Series object. This was added to `__init__.py` to ensure it points to `LuxSeries`. 

## Example Output

No example output for this PR.


dorisjlee:
Thanks @westernguy2! If I understand this correctly, we don't currently support series groupby (only `LuxDataFrameGroupBy`) but with this new change, we can now expect Lux to work with the case of series groupby? If this is the case, could you point to the specific test either on the Pandas test suite, or adding our own tests to check that series groupby is working as intended?

dorisjlee (approved):

westernguy2:
The `SeriesGroupBy` test that was failing was actually not fixed by this new implementation. However, while debugging that issue, I tried this new format for `LuxGroupBy` that seemed to be better than the old format. 

I can add a test similar to the ones related to `SeriesGroupBy` that were failing, later tonight!

westernguy2:
Also, yes, this PR also allows metadata propagation through `SeriesGroupBy` which is not supported currently, I had forgotten about that! I will add some tests for this new feature.

dorisjlee:
That's great, I was thinking that this new `SeriesGroupBy` seems to better support the metadata propagation. I'll wait for the tests before merging this in.

westernguy2:
@dorisjlee Added a test for `SeriesGroupBy`!

dorisjlee (approved):

dorisjlee:
Thanks @westernguy2, this looks good! Merging this in now!

PR #318, opened 2021-03-22
Resolve empty vis bug for geographic data types

micahtyong:
## Overview

Previously, choropleth visualizations would not appear when `x_attr` or `y_attr` was abbreviated. With this change, all choropleth visualizations should appear as expected. 

## Changes

- Modifications in `vislib/altair/Choropleth.py`

## Example Output

Now, when we run 

```
import pandas as pd
import lux
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/hpi.csv")
df
```
we should get the following output with no empty visualizations. 
Screen Shot 2021-03-22 at 3 26 52 PM

### DIFF
--- a/lux/vislib/altair/Choropleth.py
+++ b/lux/vislib/altair/Choropleth.py
@@ -69,9 +69,12 @@ class Choropleth(AltairChart):
             alt.Chart(geo_map)
             .mark_geoshape()
             .encode(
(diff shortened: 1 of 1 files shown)


dorisjlee:
This looks great, thanks @micahtyong!

PR #322, opened 2021-03-24
Support choropleths for Matplotlib backend via Altair 

### LINKED ISSUE #310: [Feature Request] Geographic data support for Matplotlib
# Feature Request Description 
Via #253, we now have support for geographic data types using Vegalite as the backend. We would like to extend support for this feature on [Matplotlib](https://matplotlib.org/) as well. 

For example, running 
```
lux.config.plotting_backend = "vegalite" # default
df
```
yields the following result: 
Screen Shot 2021-03-17 at 2 14 58 PM
When we configure the backend with
```
```
the output does not display for a Matplotlib backend. 

## Solution
To resolve for now, we will call upon the `AltairRenderer` and warn users that the Choropleths are rendered via the Altair backend. 

If users indicate enough interest, we will add a mirroring `Choropleth.py` file in `lux/vislib/matplotlib`, which will be called upon by the same `univariate` action when the backend is set to `matplotlib`.

micahtyong:
## In this PR

In this PR, we call upon the AltairRenderer and inform users that the choropleths are rendered via the Altair backend when users set their backend to Matplotlib and geographic attributes are detected. 
- If enough users indicate interest in choropleth rendering via matplotlib, we will reopen #310. 
- **Rationale**: Matplotlib doesn't support choropleths out of the box. It would require some [third party library](https://matplotlib.org/stable/thirdpartypackages/index.html#mapping-toolkits) such as Cartopy or Geopandas. While these libraries meet the necessary requirements to add Matplotlib support for choropleths, they are expensive (Cartopy is 14.6 MB, GeoPandas depends on Fiona which is 49 MB). 

## Related Issues

Closes #310. 

## Changes
- Additional code in `MatplotlibRendererer.py`
- Quick modification of repr function in `Choropleth.py` 
- Added corresponding test for choropleths in `test_vis.py`

## Example Output
Now, upon running 
```
df
```
we get the following output: 
Screen Shot 2021-03-24 at 6 38 03 PM

The output for the default Altair plotting backend remains the same.

### DIFF
--- a/lux/vislib/matplotlib/MatplotlibRenderer.py
+++ b/lux/vislib/matplotlib/MatplotlibRenderer.py
@@ -87,7 +87,7 @@ class MatplotlibRenderer:
 
             warnings.formatwarning = lux.warning_format
             warnings.warn(
-                " Choropleths ('Geographical' tab) are rendered using Altair. If you would like Matplotlib support for Choropleths, please express so on https://github.com/lux-org/lux/issues/310.\n"
+                " Choropleths ('Geographical' tab) are rendered using Altair. If you would like Matplotlib support for Choropleths, please express so on https://github.com/lux-org/lux/issues/310."
             )
             return AltairRenderer().create_vis(vis, False)
         else:
--- a/lux/vislib/matplotlib/MatplotlibRenderer.py
+++ b/lux/vislib/matplotlib/MatplotlibRenderer.py
@@ -26,7 +26,6 @@ from lux.utils.utils import matplotlib_setup
 
 import base64
 from io import BytesIO
-import warnings
 
 
 class MatplotlibRenderer:
@@ -84,10 +83,6 @@ class MatplotlibRenderer:
         elif vis.mark == "heatmap":
             chart = Heatmap(vis, fig, ax)
         elif vis.mark == "geographical":
-            warnings.formatwarning = lux.warning_format
-            warnings.warn(
-                " Choropleths ('Geographical' tab) are rendered using Altair. If you would like Matplotlib support for Choropleths, please express so on https://github.com/lux-org/lux/issues/310."
-            )
             return AltairRenderer().create_vis(vis, False)
         else:
             chart = None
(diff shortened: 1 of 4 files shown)


lux/vislib/matplotlib/MatplotlibRenderer.py
```diff
@@ -81,6 +83,12 @@ def create_vis(self, vis, standalone=True):
             chart = LineChart(vis, fig, ax)
         elif vis.mark == "heatmap":
             chart = Heatmap(vis, fig, ax)
+        elif vis.mark == "geographical":
+            warnings.formatwarning = lux.warning_format
```
dorisjlee, line 87: Could we change the warning to be displayed via `df._message.add("...msg here...")`? The issue with the regular Python "red" warnings is that it can a bit jarring to users and given that this issue is somewhat minor, it makes more sense to go into the yellow button within our widget (which is only shown when we expand the message box).

You might notice that the df is not accessible in `MatplotlibRenderer:create_vis`, so it may be possible that we have to pass the message so that it is accessible to the main dataframe (so that we can edit `df._message`).

dorisjlee (changes requested):

dorisjlee:
Thanks @micahtyong! I made some minor adjustments to the warning text and left a comment regarding moving the warning to the </img> message box instead.

lux/vislib/matplotlib/MatplotlibRenderer.py
```diff
@@ -81,6 +82,8 @@ def create_vis(self, vis, standalone=True):
             chart = LineChart(vis, fig, ax)
         elif vis.mark == "heatmap":
             chart = Heatmap(vis, fig, ax)
+        elif vis.mark == "geographical":
+            return AltairRenderer().create_vis(vis, False)
```
micahtyong, line 86: @caitlynachen I noticed you mentioned in #326 that we should avoid mixing Matplotlib and Altair, except this PR warrants that sort of use case. Do you see any issues I should look out for?

caitlynachen: ahh that was mostly taking out useless code in Matplotlib that didn't effect any of the Matplotlib rendering. In this case I think its fine to have Altair code if there's no way to render via Matplotlib.

caitlynachen (commented):
Looks great @micahtyong ! Maybe add a test to make sure the map renders the Altair Chloropleth in Matplotlib as well

dorisjlee:
This is an awesome addition to our upcoming release, thanks @micahtyong!

PR #326, opened 2021-03-26
Clean up Matplotlib code: remove altair code and replace pandas code

caitlynachen:
## Overview

1. Bug with self.code since matplotlib bar chart should not have mentions of alt.Chart . We should fix this and make sure that no other matplotlib has mentions or usage of altair stuff.
2. Inside all the matplotlib chart type,  we had originally used df = pd.DataFrame(self.data) . Is there any particular  reason why we pass self.data into a pd.Dataframe.

## Changes

1. removed altair code from matplotlib
2. use df=self.data instead of df = pd.DataFrame(self.data) 

## Example Output
matplotlib plots show as expected
Screen Shot 2021-03-25 at 8 36 35 PM

Describe the expected behavior of your changes. If applicable, please include a screenshot.


dorisjlee:
Hi @caitlynachen, Thanks for the changes. Could you resolve the merge conflicts in this PR?

dorisjlee:
Thanks @caitlynachen !

PR #331, opened 2021-03-27
Update LuxSQLTable __len__() and metadata computation""

thyneb19:
## Overview

Added __len__() function capability to the LuxSQLTable and updated metadata handling.

## Changes

Made changes to the LuxSQLTable to enable `len(sql_tbl)` functionality. Updated references to the `sql_tbl._length` parameter throughout the code to be replaced by `len(sql_tbl)`. Making these changes also required updating the metadata calculation process for the LuxSQLTable. 

During the initial setup of the LuxSQLTable, it needs to have the same `len()` functionality as the LuxDataFrame. Added a boolean parameter to keep track of which part of setup a LuxSQLTable is in, and this informs `len()` to either return `sql_tbl._length` or the parent class' `len()`.

### DIFF
--- a/lux/action/correlation.py
+++ b/lux/action/correlation.py
@@ -62,7 +62,7 @@ def correlation(ldf: LuxDataFrame, ignore_transpose: bool = True):
     }
     ignore_rec_flag = False
     # Doesn't make sense to compute correlation if less than 4 data values
-    if ldf._length < 5:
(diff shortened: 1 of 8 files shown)


dorisjlee (approved):

dorisjlee:
This looks great, thanks @thyneb19 !

PR #332, opened 2021-03-30
Fix Metadata Propagation with Overridden Pandas Methods

westernguy2:
## Overview

Currently the Pandas methods that are being overridden (`head`, `tail`, `describe`) are changing the history for the original DataFrame, instead of the newly created one. This PR fixes that issue.

## Changes

The PR moves changes the metadata of the DataFrame returned as a result of the Pandas operation. It also removes method that extends `info` since `info` doesn't return a DataFrame. The PR also ensures that the `_pandas_only` flag works in `_ipython_display_` by re-adding the else clause that was commented out.

## Example Output

The bug where two DataFrames would be outputted as a result of the lines:

```python
df.info()
df
```

was what flagged this issue, and this is now fixed.

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -624,48 +624,48 @@ class LuxDataFrame(pd.DataFrame):
             if self._pandas_only:
                 display(self.display_pandas())
                 self._pandas_only = False
-            # else:
-            if not self.index.nlevels >= 2 or self.columns.nlevels >= 2:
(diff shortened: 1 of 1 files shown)


dorisjlee (approved):

wwei76922:
6c1f6ef3703acbcab526e966f9a2e81b8708138e

PR #333, opened 2021-03-30
Add lux.config.plotting_scale factor 

### LINKED ISSUE #317: [ENH] Changing Widget Layout to enlarge chart size
Hello,

I'm not sure if this is the right place to ask.

I like Lux! Great work!

I was wondering is it possible to change the size of the widget, especially when exporting it as html? I find the height a bit small and it would be greate if you could increase the hight (and width).

I managed to increase the size of the figures but I am struggling with the size of the frame/div.
image

micahtyong:
## In this PR

Closes #317 for Altair charts in Lux. We accomplish this by adding `chart_scale` as a global property accessed and set via `lux.config.chart_scale`. 

## Changes

- New attribute in `lux/_config/config.py`
- Applying scale factor in `AltairChart.py`
- Changing `Choropleth.py` so that its height and width are defined in `AltairChart.py` 
- Applying scale factor to `mark_text` in `altair/BarChart.py` 

## Example Output

Suppose we are observing correlations within a moderately large dataset
```
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/college.csv")
df
```
sc1

The resulting output looks a bit cramped! We change this and observe the result when we prepend the above cell with 
```
lux.config.plotting_scale = 2.5 # can be replaced with any positive float
```
sc2

The result similarly applies when we set an intent. For example, 
```
df.intent = ["AverageCost","SATAverage"]
```
sc3

It's important to note that if a user tries to set the scaling factor to something other than a positive float or integer, users will see the following error
```
Scaling factor for charts must be a positive float.
```
and Lux will revert to the previously set scale factor (default is 1).

The results apply similarly for a Matplotlib backend. When we prepend the following,
```
```

Screen Shot 2021-04-19 at 5 54 35 PM

Finally, once included in the next release, this feature will come packaged with out `save_as_html()` feature.


dorisjlee:
Hi @micahtyong, These changes look great! Here's some comments: 
- This PR currently supports resizing for Altair charts, we should also incorporate support for resizing for matplotlib charts.
- I agree with the discussion in #317 that the current view visualization being smaller than the rest of the visualizations is a bit odd. For this PR,  let's make chart-scale apply to all charts. I understand that this could look a bit weird in the widget view, but since the `save_to_html` has the screen space for the whole browser, it might not be too bad. We discussed different layout options to help with this, which can be addressed in separate PR. 
- I think we have a fixed height in the current widget which means that even if we increase the scale factor, the exported HTML will still cut off the charts. We might want to address this in the layout PR mentioned above. Here is the result of the saved HTML: 
```python
lux.config.chart_scale  = 3
df = pd.read_csv("../data/car.csv")
df.save_as_html("bob.html")
```
image


@jerrysong1324: I assigned you to review the code since you might be able to reuse the `scale_factor` in the code editor pop-up that you're working on. Let us know what you think and if you have any feedback on the changes!

dorisjlee (changes requested):

jerrysong1324 (approved):

lux/vis/VisList.py
```diff
@@ -272,7 +272,13 @@ def _ipython_display_(self):
         import luxwidget
 
         recJSON = LuxDataFrame.rec_to_JSON([recommendation])
-        self._widget = luxwidget.LuxWidget(currentVis={}, recommendations=recJSON, intent="", message="")
+        self._widget = luxwidget.LuxWidget(
+            currentVis={},
+            recommendations=recJSON,
+            intent="",
+            message="",
+            plottingScale=lux.config.plotting_scale,
```
dorisjlee, line 280: Based on the frontend changes, it looks like the only usage of `plottingScale` is to increase the size of the currentVis. Or is it also being used to determine the size of the widget divs? If the former is true, can we simply increase the currentVis size like what we did inside the AltairChart and MatplotlibChart before we send it to the frontend?
That way we can minimize adding a new traitlet variable, which requires the frontend and backend to be version-locked (i.e., not backwards-compatible) for the release.

dorisjlee (changes requested):

lux/vis/VisList.py
micahtyong: Actually, it's the latter. Plotting scale is mostly used in the frontend to determine the side of the widget divs, which requires a Traitlet variable.

dorisjlee: Hmm, if we really had to pass something to the frontend, maybe we go with something more generic so that we don't have to change the spec of the traitlet too much in the future. For example, we could called the traitlet variable `config` and it would be a dictionary like `{"plotting_scale":0.5, "some other config": "..."}`. That way if we want to pass other things in the future, we can pass it along in the config. This is similar to the approach that we took for #336, so that the [`all_column` flag is lumped together with the current vis spec](https://github.com/lux-org/lux/pull/336/files#diff-b3f5355f16af7450f6bf0669466ac57ae020ed8c0a56105e8ece9c8c035a8879L699), without the need to send an additional traitlet over. We do a similar thing with `vislib` parameter too to put it alongside each vis spec.

micahtyong: Gotcha, I like the config variable idea and will include that in this PR! I agree that this is a future-proof way of more easily communicating with the frontend without additional traitlets in the future.

dorisjlee (approved):

dorisjlee (approved):

dorisjlee:
Merging this in given the test failure was caused by new traitlet.

Issue #199, opened 2021-01-06
New action: "All-column" vis when only few columns in dataframe


dorisjlee:
When there are only two columns in a dataframe involving a quantitative and a temporal attribute, we display separate univariate charts. The better chart to display here would be a line chart with both attributes. We need to change the action logic to account for this use case.

```python
tmax = pd.read_csv('https://raw.githubusercontent.com/koldunovn/python_for_geosciences/master/DelhiTmax.txt', delimiter=r"\s+", parse_dates=[[0,1,2]], header=None)
tmax.columns = ['Date', 'Temp']
```

![image](https://user-images.githubusercontent.com/5554675/103758845-bd166f00-504d-11eb-8cda-09d4c46233bd.png)

[This notebook](https://github.com/koldunovn/python_for_geosciences/blob/master/06%20-%20Time%20series%20(Pandas).ipynb) has a good use case example for time series.

dorisjlee:
A similar use case is when there are 1 dimension and 2 quantitative variables, the visualization shows as two separate charts.
![image](https://user-images.githubusercontent.com/5554675/104676717-e35d9e00-5722-11eb-8581-ab448dd2d64e.png)
![image](https://user-images.githubusercontent.com/5554675/104676725-e8225200-5722-11eb-9bf7-8d650801e749.png)
A single visualization like this would be more appropriate

![image](https://user-images.githubusercontent.com/5554675/104676820-1ef86800-5723-11eb-90e5-79a45b1bd72d.png)
![image](https://user-images.githubusercontent.com/5554675/104676837-26b80c80-5723-11eb-948d-32e9eac1d998.png)

Example code:
```python
[...]

PR #336, opened 2021-03-31
"All-column" vis when only few columns in dataframe #199

caitlynachen:
## Overview

Current-vis: "All-column" vis when only few columns in dataframe #199

## Changes

If dataframe has only 2 or 3 columns, use current_vis to show an all-column visualization

## Example Output

Screen Shot 2021-03-31 at 11 57 51 AM
Screen Shot 2021-03-31 at 11 57 35 AM

### DIFF
--- a/lux/core/frame.py
+++ b/lux/core/frame.py
@@ -90,7 +90,6 @@ class LuxDataFrame(pd.DataFrame):
         self.pre_aggregated = None
         self._type_override = {}
         warnings.formatwarning = lux.warning_format
-        self.all_column = False
 
     @property
     def _constructor(self):
@@ -346,9 +345,11 @@ class LuxDataFrame(pd.DataFrame):
             rec_infolist.append(recommendations)
 
     def show_all_column_vis(self):
-        quantitative_columns = [i for i in self.dtypes if i != "O" and i != "str"]
-        if (len(quantitative_columns) == 2 or len(quantitative_columns) == 3) and self._intent == []:
-            self.current_vis = VisList([i for i in self.columns], self)
+        if self.intent == [] or self.intent is None:
+            vis = Vis(list(self.columns), self)
+            if vis.mark != "":
+                vis._all_column = True
+                self.current_vis = VisList([vis])
 
     def maintain_recs(self, is_series="DataFrame"):
         # `rec_df` is the dataframe to generate the recommendations on
@@ -705,6 +706,10 @@ class LuxDataFrame(pd.DataFrame):
             current_vis_spec = vlist[0].to_code(language=lux.config.plotting_backend, prettyOutput=False)
         elif numVC > 1:
             pass
+        if vlist[0]._all_column:
+            current_vis_spec["allcols"] = True
+        else:
+            current_vis_spec["allcols"] = False
         return current_vis_spec
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


dorisjlee:
Hi @caitlynachen, Thanks for the PR, the all column visualization looks much better than the recommendations that we had before. This is especially true when we filter to only 2 or 3 columns in a dataframe. Here is some examples of this: 

image
image

I was playing around with this feature and had a couple suggestions on what we can do to improve this: 

- We might want to modify the text under the `Current Visualization` and title `Current Visualization` to say that the visualization is not based on the current intent, but based on all the columns in the current dataframe (`Dataframe Visualization`:`based on all columns in the dataframe`, with the text on "all columns" highlightable showing the columns involved.)
- What happens when there is an intent set and there is an actual `Current Visualization`, in this case, do we hide the all-column vis? or show it as a separate tab?
image

- You might want to look into some of the failing tests, which is related to calling the print displays

lux/core/frame.py
```diff
@@ -498,9 +498,15 @@ def maintain_recs(self, is_series="DataFrame"):
...
+            quantitative_columns = [i for i in self.dtypes if i != "O" and i != "str"]
+            if len(quantitative_columns) == 2 or len(quantitative_columns) == 3:
+                self.current_vis = VisList([i for i in self.columns], self)
             self._widget = rec_df.render_widget()
         # re-render widget for the current dataframe if previous rec is not recomputed
         elif show_prev:
+            quantitative_columns = [i for i in self.dtypes if i != "O" and i != "str"]
+            if len(quantitative_columns) == 2 or len(quantitative_columns) == 3:
+                self.current_vis = VisList([i for i in self.columns], self)
             self._widget = rec_df.render_widget()
```
jinimukh, line 510: is there a way we can write this such that we don't repeat the same code twice? You can make a quick function!

lux/core/frame.py
```diff
@@ -498,9 +498,15 @@ def maintain_recs(self, is_series="DataFrame"):
                 if len(vlist) > 0:
                     rec_df._recommendation[action_type] = vlist
             rec_df._rec_info = rec_infolist
+            quantitative_columns = [i for i in self.dtypes if i != "O" and i != "str"]
```
jinimukh, line 501: dtypes is pandas dtypes right? So what if the columns are of type DateTime, object, bools, etc? maybe make this only include i if "int" or "float" is in the type.

jinimukh (commented):
Great idea! Maybe we can extend this feature to the entire dataframe and set a global config like `small_dataframe_ok` or `show_all_data_columns` depending on the use-case you were thinking of!

lux/core/frame.py
caitlynachen: yup!

lux/core/frame.py
caitlynachen: i believe datetime works as a quantitative attr and there's lots of other types besides "int" and "float" bc bytes are specified (i think 8 total) so i figured it was easier this way

lux/core/frame.py
```diff
@@ -425,6 +426,11 @@ def _append_rec(self, rec_infolist, recommendations: Dict):
         if recommendations["collection"] is not None and len(recommendations["collection"]) > 0:
             rec_infolist.append(recommendations)
 
+    def show_all_column_vis(self):
+        quantitative_columns = [i for i in self.dtypes if i != "O" and i != "str"]
```
dorisjlee, line 430: We should probably not directly access `df.dtypes` here and use `df.data_type`

dorisjlee (changes requested):

dorisjlee (approved):

PR #337, opened 2021-04-01
Fix visualization failure when intent is missing [issue #276]

### LINKED ISSUE #276: Visualization breaks when intent is not part of resulting table
When displaying the visualization, it is possible that the intent is not present in the resulting data table. For example in the case of the index, the intent is propagated over to the resulting table, but the attribute `Attrition` is no longer in the table. When the intent is not present in the dataframe, we should handle the error by falling back to the default visualizations.

```python 
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/employee.csv")
df.intent = ["Attrition"]
df.groupby("BusinessTravel").mean()
```
image

Moh-Yakoub:
## Overview

Fixes #276 

## Changes

I have added a new method to compile safely and added fail safe getter to avoid breaking the visualization if the key doesn't exist.

## Example Output

Using missing intent doesn't break the visualization as before

Screen Shot 2021-04-01 at 8 56 23 PM

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -104,24 +104,3 @@ def test_validator_invalid_value(global_var):
         df.intent = ["Region=bob"]
 
     df.clear_intent()
-
-
-def test_validator_invalid_filter(global_var):
-    df = pytest.college_df
-
-    with pytest.raises(KeyError, match="'New England'"):
-        with pytest.warns(
-            UserWarning,
-            match="The input 'New England' looks like a value that belongs to the 'Region' attribute.",
-        ):
-            df.intent = ["New England", "Southeast", "Far West"]
-
-
-def test_validator_invalid_attribute(global_var):
-    df = pytest.college_df
-    with pytest.raises(KeyError, match="'blah'"):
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -104,24 +104,3 @@ def test_validator_invalid_value(global_var):
         df.intent = ["Region=bob"]
 
     df.clear_intent()
-
-
-def test_validator_invalid_filter(global_var):
```
dorisjlee, line 109: Is there a reason why these tests were removed?

dorisjlee (changes requested):

dorisjlee:
Hi @Moh-Yakoub, Thanks for your contribution. For the intent bug, I think once we catch the intent is breaking (or detect that the key is not in the new data columns in the Parser), we should clear the intent and show the default recommendations without the intent. @westernguy2: would love your input on how we can go about resolving this.

dorisjlee:
Addressed in #351, thanks @Moh-Yakoub!

PR #342, opened 2021-04-05
[WIP][Possible bug] Fix always true condition

Moh-Yakoub:
## Overview

I noticed that we were using a check that's likely to execute in all cases. I believe the intended behavior is to check that the number of unique vals < 20 to deduce if the column is nominal. 

The existing implementation 
`(len(ldf[attr].convert_dtypes().unique() < 20))` will return always True as long as we don't have an empty array.

I am not sure if that's the intended behavior. In case that's intended, I guess it can be written in a different way

i.e.
`(len(ldf[attr].convert_dtypes().unique() < 20)) == 0`

## Changes

I've changed the implementation to deduce the field type as nominal if the number of unique vals < 20.

## Example Output

Screen Shot 2021-04-05 at 11 53 52 PM

### DIFF
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -427,7 +427,7 @@ class PandasExecutor(Executor):
                     if (
                         convertible2int
                         and ldf.cardinality[attr] != len(ldf)
-                        and (len(ldf[attr].convert_dtypes().unique() < 20))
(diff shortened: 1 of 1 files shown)


dorisjlee:
Hi @Moh-Yakoub, This is definitely a bug in our type checking code and I've recently resolved this issue in #349. Thank you for your contribution and for raising this issue!

PR #343, opened 2021-04-05
[WIP] update type inference for string columns

### LINKED ISSUE #249: Misdetected data type when numerical column contains null
When a column contains null values, it can get misclassified as a nominal type, even if it has high cardinality and is largely quantitative. For instance, in this example, the `# Instances` and `# Attributes` are detected as `nominal` but they are better suited as `quantitative`. 

```python
import pandas as pd
import lux
df = pd.read_html("https://archive.ics.uci.edu/ml/datasets.php?format=&task=&att=&area=&numAtt=&numIns=&type=&sort=nameUp&view=table")[5]
df.columns = df.loc[0]
df = df.loc[1:]
df['Year'] = pd.to_datetime(df['Year'], format='%Y')
df.data_type
```

As a result, even when dropna occurs on these columns, the data type remains `nominal`.
```python
df[['# Instances','# Attributes']].dropna()
```
image

Moh-Yakoub:
## Overview

Closes #249

I've noticed that we have a type deduction error in case the field had NaNs as described in the attached issue.
I furthur noticed that we don't check if the string columns can be casted to a float/int type. I've added extra check to see id a string column can be casted to an int/float column and deduce the proper data_type accordingly.

Update (04/10/2021)
* Adding warning message when NaN columns displayed as histograms
* Resolve NaN series causing error in histogram `execute_binning`
* Rewrote `is_numeric_nan_column` to run 2x faster
image

## Changes

I've added a logic to 
1. Check if the column can be cast to int/double
2. apply the respective data_type inference.

## Example Output

Screen Shot 2021-04-05 at 11 51 40 PM

The result shows that the two columns mentioned in the issue: `# Instances` and `# Attributes` have a correct data_type now

### DIFF
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -19,7 +19,7 @@ from lux.core.frame import LuxDataFrame
 from lux.executor.Executor import Executor
 from lux.utils import utils
 from lux.utils.date_utils import is_datetime_series
-from lux.utils.utils import check_import_lux_widget, check_if_id_like
+from lux.utils.utils import check_import_lux_widget, check_if_id_like, return_float_or_original
 import warnings
 import lux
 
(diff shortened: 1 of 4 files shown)


dorisjlee:
Thanks @Moh-Yakoub! I made some changes to resolve the related issue in #249 and rewrite the helper function in a more optimized way. Congrats on your first contribution to Lux!

dorisjlee (approved):

PR #346, opened 2021-04-07
Matplotlib Export Bug for Scatterplot

dorisjlee:
## Overview

For matplotlib scatterplots, the figsize was inputted in the argument of `plt.subplot`, resulting in a `ValueError` when the exported code was run. I fixed this error by moving the width and height arguments to the subplot's `figsize` property.

image

image


caitlynachen (approved):
looks good!

PR #350, opened 2021-04-09
Allow different sorters to work with Vis Object

jinimukh:
## Overview

This PR introduces a way to incorporate different sorters for Vis Objects. Currently, this only includes interestingness and alphabetical sorts, but later on, we can extend this class. 

## Changes

-- add `sorters.py` in utils
-- add `lux.config.sorter` in `config`
-- add tests in `test_config`

## Example Output

TBD


lux/_config/config.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 import lux
 import warnings
+from lux.utils.sorters import Sorter
```
dorisjlee, line 9: Can we rename Sorter to something more specific? Currently, the sorter actually doesn't perform any sorting, but just returns the sort key, e.g. interestingness score, or sort key for alphabetical sorting.

lux/_config/config.py
```diff
@@ -83,6 +85,23 @@ def sort(self, flag: Union[str]):
                 stacklevel=2,
             )
 
+    @property
```
dorisjlee, line 88: Not sure if sort makes sense as a global option, since it is more common to do one type of sorting for an action, but a different sort mechanism for another action.

lux/vis/VisList.py
```diff
@@ -240,7 +240,7 @@ def sort(self, remove_invalid=True, descending=True):
         elif lux.config.sort == "descending":
             descending = True
         # sort in-place by “score” by default if available, otherwise user-specified field to sort by
-        self._collection.sort(key=lambda x: x.score, reverse=descending)
```
dorisjlee, line 243: Let's consider the case where we have a dropdown on the frontend for controlling the scoring, can we avoid regenerating the collection and simply resort based on the existing collection? This would require us to cache/store the generated collection before fetching the top K.

lux/utils/sorters.py
```diff
@@ -0,0 +1,31 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+class Sorter:
+    @staticmethod
+    def interestingness():
```
dorisjlee, line 18: Wondering if we should just merge and rename the Interestingness.py class to a Sorter. With the sorter, we should have the option to pick different interestingness metrics, such as correlation, deviation, etc (while not all of them may be applicable for different actions).

dorisjlee (changes requested):

lux/_config/config.py
jinimukh: I added two implementations -- one global ordering (this could be useful for sorting by attribute name or something really general that you want to standardize across all actions) and an action-wise ordering dictionary. If no ordering is defined for the specific action, then we fallback to the global ordering.

tests/test_config.py
```diff
@@ -295,6 +296,93 @@ def test_sort(global_var):
...
+        assert vis.get_attr_by_channel("x")[0].attribute <= string
+        string = vis.get_attr_by_channel("x")[0].attribute
+    lux.config.ordering = "y_attr"
+    df = pd.read_csv("lux/data/college.csv")
+    string = df.recommendation["Correlation"][0].get_attr_by_channel("y")[0].attribute
+    assert len(df.recommendation["Correlation"]) == 5, "Show top 5"
+    for vis in df.recommendation["Correlation"]:
+        assert vis.get_attr_by_channel("y")[0].attribute <= string
+        string = vis.get_attr_by_channel("y")[0].attribute
+    lux.config.sorter = "interestingness"
```
dorisjlee, line 322: What is the difference between `sorter` and `ordering`?

lux/utils/orderings.py
```diff
@@ -0,0 +1,76 @@
...
+        collection.sort(key=lambda x: x.get_attr_by_channel("y")[0].attribute, reverse=desc)
+
+
+def resolve_value(value):
+    if type(value) is str:
+        if value == "interestingness":
+            return Ordering.interestingness
+        elif value == "title":
+            return Ordering.title
+        elif value == "x_attr":
```
dorisjlee, line 44: can we make "title", "x_attr" and "y_attr" more readable? e.g.) "alphabetical_by_x", "alphabetical_by_y", "alphabetical_by_title"

lux/_config/config.py
dorisjlee: This sounds great, let's add some docs that explain the two different types of ways to specify sorting!

dorisjlee (changes requested):

lux/_config/config.py
```diff
@@ -137,13 +139,37 @@ def sort(self, flag: Union[str]):
...
                 "Parameter to lux.config.sort must be one of the following: 'none', 'ascending', or 'descending'.",
                 stacklevel=2,
             )
 
+    @property
+    def ordering(self):
+        return self._ordering
+
+    @ordering.setter
+    def ordering(self, value):
```
dorisjlee, line 159: Can we rename all `ordering` as `ranking`? We should avoid the word sorter in places where it might be unclear what it is referring it.

tests/test_config.py
```diff
@@ -295,6 +296,93 @@ def test_sort(global_var):
...
+    lux.config.ordering = sort_by_multiple
+    df = pd.read_csv("lux/data/college.csv")
+    df._ipython_display_()
+    cmp = (
+        df.recommendation["Correlation"][0].get_attr_by_channel("x")[0].attribute,
+        df.recommendation["Correlation"][0].get_attr_by_channel("y")[0].attribute,
+    )
+
+    for vis in df.recommendation["Correlation"]:
+        assert (
```
dorisjlee, line 348: For every assert, we should add the error message to be thrown if the assertion does not pass.

dorisjlee (changes requested):

tests/test_config.py
```diff
@@ -17,6 +17,7 @@
 import pandas as pd
 import time
 from lux.vis.VisList import VisList
+from lux.utils.orderings import Ordering
```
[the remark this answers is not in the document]
jinimukh: I'll go for ranking because sort order (i.e. ascending/descending) uses "sort".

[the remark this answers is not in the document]
dorisjlee: ok!

PR #351, opened 2021-04-09
Fix error when intent is invalid

### LINKED ISSUE #276: Visualization breaks when intent is not part of resulting table

westernguy2:
## Overview

Should fix #276. Checks if the intent is in the dataset, and if not, doesn't use that intent in compilation and the creation of `current_vis`

## Changes

This PR filters through each attribute in the intent to ensure it's in the dataset, and disregards any attribute not in the dataset.

## Example Output

The example in the issue linked above shouldn't error anymore.


lux/core/frame.py
```diff
@@ -305,12 +305,28 @@ def recommendation(self, recommendation: Dict):
...
             self._current_vis is not None
             and len(self._current_vis) > 0
             and self._current_vis[0].data is None
-        ):
-            lux.config.executor.execute(self._current_vis, self)
+            and self._current_vis[0].intent
+        )
+        if valid_current_vis:
+            for clause in self._current_vis[0].intent:
+                if isinstance(clause.attribute, list):
```
dorisjlee, line 335: Can you explain the logic of the checks that are performed here? Is there a reason why we can't simply use the same validation logic inside the Parser and Validator?

lux/processor/Compiler.py
```diff
@@ -82,7 +82,21 @@ def compile_intent(ldf: LuxDataFrame, _inferred_intent: List[Clause]) -> VisList
         vis_collection: list[lux.Vis]
                 vis list with compiled lux.Vis objects.
         """
-        if _inferred_intent:
+        valid_intent = _inferred_intent  # ensures intent is non-empty
```
dorisjlee, line 85: Same comments as above with this. Are these the same checks? Can we share these?

lux/core/frame.py
westernguy2: The check here just checks if the `clause.attribute` is "?", "Record", or an attribute in the columns of the DataFrame (both in the case that the `clause.attribute` is a String or a Python list.

I will look into using `validate_intent` to see if we can use some of the same logic here instead!

lux/processor/Compiler.py
westernguy2: same as above!

Issue #240, opened 2021-01-23
Extend support for ordinal data type


dorisjlee:
Ordinal data are common in rating scales for surveys, as well as attributes like Age. 
Ordinal data currently gets classified as categorical, especially if the column contains NaN values.
The [young people survey dataset](https://www.kaggle.com/miroslavsabo/young-people-survey) on Kaggle is a good example of this, since it contains lots of rating scale data. 
![image](https://user-images.githubusercontent.com/5554675/105573876-805fad00-5d9b-11eb-8448-49bf0e8b8fe5.png)
This issue should extend support for ordinal data type detection, as well as better visualizations to display for ordinal data type. For example, ordinal data bar charts should be ordered instead of sorted based on the measure values. In addition, correlation of one or more ordinal attribute would be relevant to show.

dorisjlee:
The absenteeism dataset actually has a couple very interesting columns (e.g., Body mass index, Height)  that are quantitative but due to the integer nature and low~medium cardinality, it is detected as nominal. I'm wondering if this would actually be a good use case for the ordinal data type as some intermediate in between. In particular, I feel that nominal is especially inappropriate since we would ideally want a scatterplot for something like BMI and not have these columns be part of Filters with equalities. 
```python
df = pd.read_csv("../lux-datasets/data/absenteeism.csv")
df.intent = ["Weight"]
df
```
![image.png](https://images.zenhubusercontent.com/60014d8bd883b389bb58ec9c/a4e1770c-5596-41f8-a93e-050946a774b2)

PR #360, opened 2021-04-15
Ordinal Data Type

jinimukh:
## Overview

This PR addresses #240 by adding support for the ordinal data type. Currently, the only way to set the data type to ordinal is by using `df.set_data_type({"col_name": "ordinal})` functionality.  Optionally, if the entries do not have a natural ordering like number or alphabetical, a custom ordering can be specified using `df.set_data_type({"col_name": "ordinal}, order={"col_name": [ordered_lst]})`. To visualize ordinal data types, we are using boxplots but because they are bivariate distributions, they only show up to enhance a selected visualization. 

## Changes

- `univariate.py`: allow `ordinal` data types to be treated as `nominal` data types to create bar graphs in `Occurrences` tab
- `frame.py`: allow the `set_data_type` function to take in optional `order` argument to specify orders on ordinal data
- `BoxPlot.py`: currently only supports Altair BoxPlots
- `Compiler.py`: allow the mark to be `box` when `n_dim == 1 and n_msr == 1 and `dimension_type == "ordinal"`

## Example Output

Screen Shot 2021-04-14 at 1 51 27 PM


lux/vislib/altair/AltairRenderer.py
```diff
@@ -21,6 +21,7 @@
 from lux.vislib.altair.Histogram import Histogram
 from lux.vislib.altair.Heatmap import Heatmap
 from lux.vislib.altair.Choropleth import Choropleth
+from lux.vislib.altair.BoxPlot import BoxPlot
```
dorisjlee, line 24: If the user uses matplotlib for boxplots, could we render the boxplot in Altair and show an info button message letting users know that the matplotlib boxplot is not currently implemented? This is similar to what we did for the geographical maps in matplotlib.

lux/action/univariate.py
```diff
@@ -63,7 +63,10 @@ def univariate(ldf, *args):
             ignore_rec_flag = True
     elif data_type_constraint == "nominal":
         possible_attributes = [
-            c for c in ldf.columns if ldf.data_type[c] == "nominal" and c != "Number of Records"
+            c
```
dorisjlee, line 66: This line split is pretty weird and hard to read. Can we fix this and add a comment on what this list of possible_attributes is used for?

lux/vislib/altair/AltairRenderer.py
jinimukh: I've implemented the Altair fallback as well as the message. However, since I'm not being able to set intent on the dataframe due to the matplotlib bug, I'm not sure if the message works. Let me know if you'd like me to remove it since there is no way to verify!

dorisjlee:
Can we file a follow-up issue to delegate boxplot calculations to the Pandas and SQL Executor? This will help with performance by bringing down the rendering speed from the cost of a scatterplot to that of a boxplot (several summary statistics + outliers).

dorisjlee:
I'm wondering if ordinal data types have to be a subset of nominal data? Apart from the documentation and within the actions logic (`enhance` and `univariate`), is there anything in the code that treats ordinal as a subset of nominal. For example, can we capture scenarios where ordinal data type could be a subset of temporal data type? Such as `{Summer, Winter, Fall}`, `{Q1, Q2, …}`. It would be helpful to add an example for this.

dorisjlee:
Here's some examples that I was playing around with: 
```python
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/aug_test.csv")
df =df.dropna(subset=['education_level',"company_size"])
df.set_data_type({'education_level': "ordinal"}, 
                 order={'education_level': ['Primary School', 'High School', 'Masters','Graduate', 'Phd']})
df["education_level"]


df.set_data_type({'company_size': "ordinal"}, 
                 order={'company_size': [
                     '<10', '10/49', '50-99', '100-500',
                       '500-999', '1000-4999', '5000-9999','10000+'
                 ]})
df["company_size"]
```

I was initially a bit confused by why the boxplot was not shown for the number of records case in univariate (until we set the intent), then I realized that the boxplot didn't make sense for the ordinal data type. I wonder if it makes sense to have a bivariate ordinal data type tab, i.e., ordinal with respect to all measure values, so that the boxplot could be shown in the initial view.
Otherwise, it would appear that setting the intent doesn't change anything.

jinimukh:
> I'm wondering if ordinal data types have to be a subset of nominal data? Apart from the documentation and within the actions logic (`enhance` and `univariate`), is there anything in the code that treats ordinal as a subset of nominal. For example, can we capture scenarios where ordinal data type could be a subset of temporal data type? Such as `{Summer, Winter, Fall}`, `{Q1, Q2, …}`. It would be helpful to add an example for this.

PR #363, opened 2021-04-15
Update to executor documentation.

thyneb19:
## Overview

Update to executor documentation.


doc/source/advanced/executor.rst
```diff
@@ -2,16 +2,26 @@
...
 
-
-
 SQL Executor
 =============
 
-Lux extends its visualization exploration operations to data within SQL databases. By using the SQL Executor, users can specify a SQL database to connect a LuxSQLTable for generating all the visualizations recommended in Lux.
+Lux extends its visualization capabilities to SQL within Postgres databases. By using the SQLExecutor, users can create a :code:`LuxSQLTable` that connects to a Postgres database. When the :code:`LuxSQLTable` object is printed out, Lux displays a subset of the data and recommends a default set of visualizations to display.
+
+.. image:: https://github.com/lux-org/lux/blob/sql-engine/doc/source/img/SQLexecutor.gif?raw=true
```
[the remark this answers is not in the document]
dorisjlee: Also the GIF seems to be based on an old version of the frontend, since the preview caption and buttons are different. Can we do a new recording of the GIF with several changes:
1) Update the frontend to be based on latest version of code
2) We don't need to show the data upload cell, it would be good to just zoom in to the statement on where the LuxSQLTable is created. Something similar to this, where the import, pycopg2 connection, and LuxSQLTable declaration is separated into three cells:
```python
import lux
import pandas as pd
```

```python
import psycopg2
connection = psycopg2.connect("host=localhost dbname=testdb user=testuser password=testpass")
lux.config.set_SQL_connection(connection)
```


```python
tbl = lux.LuxSQLTable("college")
tbl
```
image
3) The GIF doesn't need to spend a lot of time scrolling through and clicking on the tabs of the widget, simply Toggle to the Lux view and clicking on one tab would be sufficient.

doc/source/advanced/executor.rst
```diff
@@ -22,20 +32,20 @@ To do this, users first need to specify a connection to their SQL database. This
 .. code-block:: python
 
 	import psycopg2
-	connection = psycopg2.connect("dbname=example_database user=example_user, password=example_password")
+	connection = psycopg2.connect("dbname=postgres_db_name user=example_user password=example_user_password")
```
dorisjlee, line 36: Should we also give an example here for sqlalchemy and say that it also works with sqlalchemy? Users are also responsible for installing psycopg2 or sqlalchemy on their own.

dorisjlee (changes requested):

doc/source/advanced/executor.rst
thyneb19: I think that would make sense, can add a quick setup of the sqlalchemy engine just to give an example to users.

dorisjlee (approved):

PR #368, opened 2021-04-27
Scalability: incorporate early pruning optimizations

dorisjlee:
## Overview

Incorporating early pruning optimizations described in [our recent paper](http://dorisjunglinlee.com/files/lux-vldb-tr.pdf) to speed up cases where the visualization search space is large (e.g., VisList contains more than top K visualizations and the dataframe has a large number of rows).

## Changes

- Adding in `execute_approx_sample` for approximating samples for early pruning
- Apply early pruning based on empirically-tested width and length criteria
- Added tests and warning messages when early pruning conditions are met

Others: 
- Increasing the sampling start and cap for overall sampling
- Adding config flag to turn lazy maintenance on/off
- Added cached method for `df.unique()`

## Example Output

Here are some results from preliminary experiments on the 1M Airbnb dataset. We duplicate a single visualization N number of times and measuring the performance. The time is the sum of the time it takes to go through the search space and the time that it takes to recompute the top K visualization upon display of the VisList. The significant speedup in the former time supersedes extra time it takes for recompute, around 17, typically no more than a few more than 15 (which k=15).
<div>
    </img>
    
    </img>
</div>

Zooming out, the speedup can be significant when searching through over 100+ vis: 
    </img>

### DIFF
--- a/lux/_config/config.py
+++ b/lux/_config/config.py
@@ -25,16 +25,23 @@ class Config:
         # flags whether or not an action has been registered or removed and should be re-rendered by frame.py
         self.update_actions: Dict[str, bool] = {}
         self.update_actions["flag"] = False
-        self._sampling_start = 10000
-        self._sampling_cap = 30000
-        self._sampling_flag = True
-        self._heatmap_flag = True
         self._plotting_backend = "vegalite"
         self._topk = 15
         self._sort = "descending"
         self._pandas_fallback = True
         self._interestingness_fallback = True
         self.heatmap_bin_size = 40
+        #####################################
+        #### Optimization Configurations ####
+        #####################################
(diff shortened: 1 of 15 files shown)


thyneb19 (approved):
Hey Doris the changes overall look good, also thanks for doing the quick fix for the SQLExecutor. I think there is more work that needs to be done to make the pruning actually compatible with the SQLExecutor case, but maybe we can merge in the PandasExecutor implementation first.

dorisjlee:
Thanks @thyneb19! There definitely needs to be more work to support pruning in SQLExecutor in the future, which would require adding an implementation for `execute_approx_sample`. Will merge this in for now!

PR #379, opened 2021-05-23
Documentation for lux.config.plotting_scale

micahtyong:
## In this PR

Update documentation for adjusting a chart's [plotting scale](https://github.com/lux-org/lux/pull/333). 

## Changes

- New section for plotting scale in `doc/source/reference/config.rst`

## Example Output

Screen Shot 2021-05-23 at 1 16 52 PM


doc/source/reference/config.rst
```diff
@@ -65,6 +65,31 @@ If you try to set the default_display to anything other than 'lux' or 'pandas,'
   :width: 700
   :align: center
 
+
+Change the plotting scale 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We can set the :code:`plotting_scale` to scale the size of charts displayed in the Lux widget. In the following block, we set plotting scale to 2, therefore the Lux widget will display charts two times larger than the default.
```
dorisjlee, line 72: Some suggested changes: 
```
We can adjust the sizes of charts displayed in the Lux widget based on a :code:`plotting_scale` applied to all charts. The default plotting scale is set to 1, displayed as followed: 
[add screenshot of original widget]
In the following block, we set the :code:`plotting_scale` to 2. We see that the Lux widget displays charts two times larger than the default. 
```

doc/source/reference/config.rst
```diff
@@ -65,6 +65,31 @@ If you try to set the default_display to anything other than 'lux' or 'pandas,'
...
+.. code-block:: python
+
+    lux.config.plotting_scale = 2 
+    df
+
+.. image:: https://github.com/lux-org/lux-resources/blob/master/doc_img/scale-1.png?raw=true
+  :width: 700
+  :align: center
+
+When specifiying intent, the plotting scale applies to both the current visualization and the chart gallery.
```
dorisjlee, line 83:
```
When further operations are performed (e.g., specifying intent), the user-defined plotting scale continues to apply to both the current visualization and recommended charts. 
```

doc/source/reference/config.rst
dorisjlee: I suggest that we add the original widget so that we can see the sizes for comparison.

doc/source/reference/config.rst
dorisjlee: This statement and image about specifying intent are not entirely necessary, since there is nothing special about specifying intent from the user's point of view. Instead, it might be cool to showcase the shrinking example.
```
In addition to specifying :code:`plotting_scale` as a positive integer (i.e., >1, enlarging the charts), :code:`plotting_scale` can also take in any positive float between [0,1] to shrink the size of the charts.
[add example]
lux.config.plotting_scale = 0.5 
```

dorisjlee (changes requested):

dorisjlee (approved):

dorisjlee (approved):

dorisjlee:
Thanks @micahtyong !

PR #388, opened 2021-06-23
Update to SQLExecutor and Visualization Code Exporting

thyneb19:
## Overview

Changed the SQLExecutor to use the more generalized template system. Instead of only working with Postgresql connections, the new SQLExecutor can be made compatible with a wider set of relational database systems as long as the appropriate query template file is provided. Currently we have also included templates for Postgres and MySQL databases.

Python code exporting has also been updated. The formatting has now been fixed using autopep8, and exported Lux Vis code can now be run without having to manually fix the code. In addition, a test suite has been added to ensure that the exported Vis code is executable.
## Changes

Update to the SQLExecutor, now takes on the structure of the GeneralizedDBExecutor that we used to have. Also updated the code formatting for the Pandas and SQL Executors. This formatting removed line wrapping which helps clean up Lux code when exported.

Added test suite for both executors under the test_export.py file. In addition, the black formatting rules have been updated to ignore the Executor files and leave their code lines unwrapped.


lux/_config/config.py
```diff
@@ -377,11 +382,23 @@ def set_SQL_connection(self, connection):
...
+                query_dict[key] = val.strip()
+        self.query_templates = query_dict
+        self.executor = SQLExecutor()
+
     def set_executor_type(self, exe):
         if exe == "SQL":
             from lux.executor.SQLExecutor import SQLExecutor
 
             self.executor = SQLExecutor()
+            self.read_query_template("../lux/lux/_config/postgres_query_template.txt")
```
dorisjlee, line 401: This seems to be an absolute path. When I tried it out the template file could not be found.
Screen Shot 2021-06-23 at 5 00 40 PM

lux/core/joinedsqltable.py
```diff
@@ -0,0 +1,229 @@
...
+
+# from lux.executor.Executor import *
+import warnings
+import traceback
+import lux
+
+
+class JoinedSQLTable(lux.LuxSQLTable):
+    """
+    A subclass of Lux.LuxDataFrame that houses other variables and functions for generating visual recommendations. Does not support normal pandas functionality.
```
dorisjlee, line 34: Could we add an explanation/example of how JoinedSQLTable can be used?

Issue #408, opened 2021-08-16
[BUG] Lux showing only max 10 values on x-axis. How to increase it?


SnowRipple:
Sorry, not sure if this is a bug, but I cannot find the information anywhere.

When visualizing more than 10 values I get (" + X more ...") in the top right corner but I cannot "unroll it". Seems like Lux is limited to just 10 values on X-axis, how do it increase this limit?

dorisjlee:
Hi @SnowRipple, We have added an additional configuration to change this default. See [these docs for more details](https://lux-api.readthedocs.io/en/latest/source/reference/config.html#change-the-maximum-number-of-bars-displayed). This will be updated along with our next release, thanks!

PR #414, opened 2021-08-22
Additional configuration options for number of bars and axis labels

adityagp:
## Overview

Resolves issue #408. In addition adds the ability to set display axis label lengths.
 
## Changes

Changes made to altair and matplotlib bar chart specifications, as well as configuration. 

## Example Output

Should allow users to set number of bars in bar charts, and change default max axis label lengths.


dorisjlee:
This looks great, I've added some docs for the new config. Thanks, @adityagp!

Screen Shot 2021-08-24 at 10 16 45 AM

Issue #398, opened 2021-07-05
[BUG] LuxDataFrame incompatible with latest pandas 1.3.0


dorisjlee:
**Describe the bug**
LuxDataFrame wrapper not working for the [latest pandas version 1.3.0](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.3.0.html#what-s-new-in-1-3-0-july-2-2021). 
pd.read_csv results in a Pandas dataframe, instead of a Lux dataframe, causing all the custom Lux methods to fail when called. This is a trace of the error from [the Github action](https://github.com/lux-org/lux/runs/2992329961#step:8:7477).

A temporary fix has been made to cap the pandas version to below 1.3

**To Reproduce**
```
pip install --upgrade pandas
```
This upgrades the pandas to the latest version. 
```python
import lux
import pandas
df = pd.read_csv("https://github.com/lux-org/lux-datasets/blob/master/data/car.csv?raw=true")
type(df)
> pandas.core.frame.DataFrame
```

dorisjlee:
Resolved in #428

PR #428, opened 2021-10-27
FIX-#398: Add fix for pandas 1.3

westernguy2:
Signed-off-by: Kunal Agarwal <kagarwal2@berkeley.edu>

## Overview

Pandas 1.3.x changes the file structure of `pd.io.parsers`. As a result, in order to read any data in as a LuxDataFrame, the following change must be made.

## Changes

I've changed `__init__` to account for the new file structure mentioned above. I've also updated the requirements to require `pandas>= 1.3.0` since the change will cause Lux to break on any Pandas version below 1.3.0.

Currently 3 tests are failing due to a different update in Pandas 1.3.x which causing history to not propagate in some instances.

## Example Output

The example given in #398 displays the Lux widget as expected now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 scipy>=1.3.3
 altair>=4.0.0
 numpy>=1.16.5
-pandas>=1.2.0,<1.3.0
+pandas>=1.3.0
 scikit-learn>=0.22
 matplotlib>=3.0.0
 lux-widget>=0.1.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 scipy>=1.3.3
 altair>=4.0.0
 numpy>=1.16.5
-pandas>=1.3.0
+pandas
 scikit-learn>=0.22
 matplotlib>=3.0.0
 lux-widget>=0.1.4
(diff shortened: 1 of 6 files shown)


dorisjlee (approved):

dorisjlee:
This looks great, thanks @westernguy2!
(Note to future dev: some metadata propagation related tests are commented out here and would need more investigation for pandas> v1.3 )

requirements.txt
```diff
@@ -1,7 +1,7 @@
 scipy>=1.3.3
 altair>=4.0.0
 numpy>=1.16.5
-pandas>=1.2.0,<1.3.0
+pandas
```
dhirschfeld, line 4: > *I've also updated the requirements to require `pandas>= 1.3.0`*

There seems to be no minimum bound here?

dorisjlee: I agree that enforcing the pandas>=1.3.0 would be useful here, we originally took this out because we didn't want to force users into updating their pandas version from 1.2 to 1.3 because of their updates to Lux. 

@westernguy2, @cgarciae : Would love to get your thoughts on this! With this PR fix, my understanding is that we no longer support pandas<1.3? Does using pandas 1.2 cause errors in Lux or fail silently?

westernguy2: The PR fix does now support the most up to date version of `pandas` and `pandas<1.3`. In `lux/core/__init__`, I check for the version, so the correct variable assignment is completed based on the version. 

`pandas 1.2` should work as expected

dorisjlee: Got it, thanks @westernguy2 !

Issue #431, opened 2021-11-10
[BUG] Issues with Sampling Start and Cap


westernguy2:
**Describe the bug**
DataFrames that are of size slightly below the sampling cap will end up having a larger sample of the data than DataFrames that are slightly above the sampling cap. This shouldn't be happening, as the size of the sample of the DataFrame should grow with the size of the DataFrame.
**To Reproduce**
The description of the bug is a clear explanation to reproduce the bug.

**Expected behavior**
This is also described in the bug description above.

PR #432, opened 2021-11-10
FIX-#431: implement sampling threshold and edit tests and docs

westernguy2:
Signed-off-by: Kunal Agarwal <kagarwal2@berkeley.edu>

## Overview

I removed the sampling_cap and sampling_start from the config and `execute_sampling`. Instead, I added a `sampling_thresh`, which is a threshold for which we'd begin sampling. This threshold forces any data larger than the threshold to sample the data such that the sample is equal to the size specified by the threshold. More information can be found in the documentation that was edited as part of this PR.

## Changes

I edited `config` and `execute_sampling` to implement this fix. I also edited the tests to account for the removal of the old configs and the addition of the new one. I finally edited the documentation to reflect these changes.

## Example Output

The issue described in #431 should be fixed now.

### DIFF
--- a/lux/_config/config.py
+++ b/lux/_config/config.py
@@ -44,8 +44,7 @@ class Config:
         #####################################
         #### Optimization Configurations ####
         #####################################
-        self._sampling_start = 100000
-        self._sampling_cap = 1000000
+        self._sampling_thresh = 100000
         self._sampling_flag = True
         self._heatmap_flag = True
         self._heatmap_start = 5000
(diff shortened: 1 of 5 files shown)


lux/_config/config.py
```diff
@@ -186,55 +185,27 @@ def interestingness_fallback(self, fallback: bool) -> None:
...
-            )
-
-    @property
-    def sampling_start(self):
+    def sampling_thresh(self):
         """
         Parameters
         ----------
         sample_number : int
             Number of rows required to begin sampling. Must be smaller or equal to _sampling_cap
```
dorisjlee, line 193: change description in docstring

tests/test_maintainence.py
```diff
@@ -96,9 +98,10 @@ def test_intent_cleared_after_vis_data():
         lux.Clause("City=Crofton"),
     ]
     df._ipython_display_()
-
     vis = df.recommendation["Similarity"][0]
     vis.data._ipython_display_()
     all_column_vis = vis.data.current_vis[0]
     assert all_column_vis.get_attr_by_channel("x")[0].attribute == "Year"
+    print(all_column_vis.get_attr_by_channel("y"))
```
dorisjlee, line 105: remove print statement

dorisjlee (approved):

dorisjlee:
The code changes looks great. It seems like `test_maintainence.py` is failing because the 10k row sample is probably not sufficient for these use cases. The sampling can be problematic for Similarity or Filter based actions. In this case, given that the intent involves filter, the Similarity action iterates over all the line charts on each filter.

PR #438, opened 2021-12-01
FIX-#431: Moving and adding sampling to backend calculations

westernguy2:
## Overview

This is a branch that builds off of the work done in #432. This moves the sampling to after the Filter and also adds sampling for metadata computation.

## Changes

Changes the `execute` function to move sampling after the Filtering is done, and so the sampling is done on each of the data visualizations. It also edits `compute_data` to sample the data before computing the metadata. All metadata is the metadata associated with the sample, not the full dataset.

## Example Output

N/A

### DIFF
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -57,13 +57,14 @@ class PandasExecutor(Executor):
 
         if SAMPLE_FLAG and len(ldf) > SAMPLE_THRESH:
             if ldf._sampled is None:  # memoize unfiltered sample df
-                ldf._sampled = ldf.sample(n=SAMPLE_THRESH, random_state=1)
+                final_df = ldf.sample(n=SAMPLE_THRESH, random_state=1)
             ldf._message.add_unique(
                 f"Large dataframe detected: Lux is only visualizing a sample of {SAMPLE_THRESH} rows.",
                 priority=99,
             )
         else:
-            ldf._sampled = ldf
+            final_df = ldf
+        return final_df
 
     @staticmethod
     def execute_approx_sample(ldf: LuxDataFrame):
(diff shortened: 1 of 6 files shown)


lux/executor/PandasExecutor.py
```diff
@@ -113,28 +106,28 @@ def execute(vislist: VisList, ldf: LuxDataFrame, approx=False):
         None
         """
 
-        PandasExecutor.execute_sampling(ldf)
+        filter_executed = {}
```
dorisjlee, line 109: Not sure if I understand why we need to store `filter_executed` as a dictionary and what its function is.

lux/executor/PandasExecutor.py
```diff
@@ -443,16 +436,16 @@ def compute_data_type(self, ldf: LuxDataFrame):
                     ldf._data_type[attr] = "geographical"
                 elif pd.api.types.is_float_dtype(ldf.dtypes[attr]):
 
-                    if ldf.cardinality[attr] != len(ldf) and (ldf.cardinality[attr] < 20):
+                    if ldf.cardinality[attr] != ldf._length and (ldf.cardinality[attr] < 20):
```
dorisjlee, line 439: What is the difference between `_length` and `len(df)`? It is probably more general to use the latter since the `_length` might not be maintained correctly.

lux/executor/PandasExecutor.py
```diff
@@ -538,11 +531,17 @@ def _is_datetime_number(series):
         return False
 
     def compute_stats(self, ldf: LuxDataFrame):
+        # use sample to compute statistics
+        if ldf._sampled is None:
+            ldf_sampled = PandasExecutor.execute_sampling(ldf)
```
dorisjlee, line 536: Will the config parameters that we are using for sampling for metadata and the visualization be the same?

dorisjlee (changes requested):

lux/executor/PandasExecutor.py
westernguy2: Yes, this was some old code when I had split this into two different for loops. I'll put it back to using just a variable to save it.

lux/executor/PandasExecutor.py
westernguy2: Yes, currently they are the same (`sampling_thresh`). Should we maybe use different parameters?

lux/executor/PandasExecutor.py
westernguy2: I noticed `_length` in the metadata for a LuxDataFrame, and found that it was not being used anywhere in the code base (as far as I could tell). On Line 544 of this file, I changed it to be the length of the sampled DataFrame. This is necessary since we don't save the sampled DataFrame after the metadata is computed, but the length of the sampled DataFrame is necessary for future calculations, especially ones related to cardinality, like the one here.

The name of the attribute is probably not the best, so I could maybe change it to `_sampled_length` instead?

Issue #417, opened 2021-09-11
[BUG] AttributeError: 'LuxDataFrame' object has no attribute '_pandas_only'


Pisciotta:
After loading a particular dataframe from pickle using both pandas.read_pickle or pickle.load, I get the following error 

C:\Anaconda3\envs\...\lib\site-packages\IPython\core\formatters.py:918: UserWarning:
Unexpected error in rendering Lux widget and recommendations. Falling back to Pandas display.
Please report the following issue on Github: https://github.com/lux-org/lux/issues 

C:\Anaconda3\envs\...\lib\site-packages\lux\core\frame.py:632: UserWarning:Traceback (most recent call last):
  File "C:\Anaconda3\envs\featex\lib\site-packages\lux\core\frame.py", line 576, in _ipython_display_
    if self._pandas_only:
  File "C:\Anaconda3\envs\featex\lib\site-packages\lux\core\frame.py", line 177, in __getattr__
    ret_value = super(LuxDataFrame, self).__getattr__(name)
  File "C:\Anaconda3\envs\featex\lib\site-packages\pandas\core\generic.py", line 5465, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'LuxDataFrame' object has no attribute '_pandas_only'

The problem must be related to my particular dataframe since Lux is working properly (with just a Warning) if I use some standard dataframes (e.g. from Seaborn). Still, I do not understand the problem because it is being displayed normally from the Pandas display. Any idea about that? Thank you in advance.

PR #445, opened 2022-01-13
FIX-#417: Add test for datetime index

cgarciae:
## Overview

Issue #417 suggest datetime index after de-serialization might cause issues, however the error was not reproducible by the added tests.

## Changes
* Adds two tests with dataframes that have datetime indexes.

### DIFF
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -86,7 +86,8 @@ def test_datetime_index():
     # continuous
     c1 = np.random.uniform(0, 1, size=nrows)
 
-    df = pd.DataFrame({"c1": c1}, index=dt)
+    data = np.random.uniform(0, 1, size=(nrows,50))
+    df = pd.DataFrame(data, index=dt)
 
     df._ipython_display_()
 
--- a/tests/test_pandas.py
+++ b/tests/test_pandas.py
@@ -78,7 +78,7 @@ def test_infs():
 
 
 def test_datetime_index():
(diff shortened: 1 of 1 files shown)


tests/test_pandas.py
```diff
@@ -73,3 +75,36 @@ def test_infs():
...
+def test_datetime_index():
+    nrows = 100000
+
+    # create a datetime index, freq in seconds
+    dt = pd.date_range("1/1/2019", periods=nrows, freq="1s")
+
+    # continuous
+    c1 = np.random.uniform(0, 1, size=nrows)
+
+    df = pd.DataFrame({"c1": c1}, index=dt)
```
dorisjlee, line 89:
```suggestion
    data = np.random.uniform(0, 1, size=(nrows,100))
    df = pd.DataFrame(data, index=dt)
```

dorisjlee: Tested a larger dataframe here and there's no error. Also tested the transpose `df.T` which leads to some warnings but no error.

tests/test_pandas.py
```diff
@@ -73,3 +75,36 @@ def test_infs():
     df = pd.DataFrame({"c1": c1, "c2": c2, "d1": d1, "d2": d2})
 
     df._ipython_display_()
+
+
+def test_datetime_index():
+    nrows = 100000
```
dorisjlee, line 81:
```suggestion
    nrows = 100
```

PR #446, opened 2022-01-18
FIX-#415: Default to Sturges' rule for histograms with few data points

### LINKED ISSUE #415: [BUG] Bin Determination Edge Case
**Describe the bug**

In the college dataset, there are only 23 rows where HighestDegree='Associate', as a result, the bin size determined via the new formula (#302) is incorrect. 

**To Reproduce**

```python
import lux
import pandas as pd
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/college.csv")
df.intent = ["SATAverage","HighestDegree=Graduate"]
df
```
Screen Shot 2021-08-26 at 4 29 56 PM

micahtyong:
## Overview

Resolves #415 by defaulting to Sturges' rule to handle histograms with few data points (which results in histogram bars that are too skinny / too fat). 

## Changes

Modified edge case handling in `altair/Histogram.py`. Behavior for non-edge cases still uses Freedman Diaconis's Rule and remains unchanged.

## Example Output

Upon running 

```
df = pd.read_csv("https://raw.githubusercontent.com/lux-org/lux-datasets/master/data/college.csv")
df.intent = ["SATAverage","HighestDegree=Graduate"]
df
```

and clicking the "Filter" tab, we get the following widget displayed: 

Screen Shot 2022-01-17 at 6 57 04 PM


dorisjlee (approved):

dorisjlee:
This looks great, thanks @micahtyong!

Issue #403, opened 2021-07-30
Excel file error with `pd.cut`


METTA2:
**Describe the bug**

I am excited to use Lux but am frustrated by the following error:

**C:\Users\Michael\miniconda2\lib\site-packages\IPython\core\formatters.py:918: UserWarning:
Unexpected error in rendering Lux widget and recommendations. Falling back to Pandas display.
Please report the following issue on Github: https://github.com/lux-org/lux/issues 

C:\Users\Michael\miniconda2\lib\site-packages\lux\core\frame.py:628: UserWarning:Traceback (most recent call last):
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\core\frame.py", line 590, in _ipython_display_
    self.maintain_recs()
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\core\frame.py", line 432, in maintain_recs
    custom_action_collection = custom_actions(rec_df)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\action\custom.py", line 76, in custom_actions
    recommendation = lux.config.actions[action_name].action(ldf)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\action\correlation.py", line 50, in correlation
    vlist = VisList(intent, ldf)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\vis\VisList.py", line 43, in __init__
    self.refresh_source(self._source)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\vis\VisList.py", line 336, in refresh_source
    lux.config.executor.execute(self._collection, ldf, approx=approx)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\executor\PandasExecutor.py", line 143, in execute
    PandasExecutor.execute_2D_binning(vis)
  File "C:\Users\Michael\miniconda2\lib\site-packages\lux\executor\PandasExecutor.py", line 422, in execute_2D_binning
    vis._vis_data["yBin"] = pd.cut(vis._vis_data[y_attr], bins=lux.config.heatmap_bin_size)
  File "C:\Users\Michael\miniconda2\lib\site-packages\pandas\core\reshape\tile.py", line 238, in cut
    rng = (nanops.nanmin(x), nanops.nanmax(x))
  File "C:\Users\Michael\miniconda2\lib\site-packages\pandas\core\nanops.py", line 135, in f
    result = alt(values, axis=axis, skipna=skipna, **kwds)
[...]

Issue #394, opened 2021-06-29
Mixed dtype data causes error in `pd.cut`


capnahab:
I am not experienced in data analysis, - I am a surgeon looking at operation data for pituitary surgery. I am using a macbook pro with apple M 'silicon' processor.

Lux will not load after importing a csv in jupyter notebook. The csv is 62 fields and 6000 rows but fairly simple with some datetimes in about 6 fields. Defaults back to Pandas. 

I have the jupyter nbextension luxwidget in my extensions tab and no problems reported on 
jupyter nbextension install --py luxwidget
jupyter nbextension enable --py luxwidget

to reproduce. 

`import` pandas as pd
import lux

df = pd.read_csv("/Users/Desktop/ND/June_20/data/with.csv")
df[['ADMIDATE', 'DISDATE', 'EPIEND', 'EPISTART', 'OPDATE_01','DOD']] = df[['ADMIDATE', 'DISDATE', 'EPIEND', 'EPISTART', 'OPDATE_01','DOD']].apply(pd.to_datetime, dayfirst=True, errors='coerce')
df['fyear'] = pd.to_datetime(df['fyear'], format='%Y')

pit = ["B041", "B012"]
df_p = df[df["OPERTN_01"].isin(pit)].copy()
df_p

/opt/homebrew/lib/python3.9/site-packages/IPython/core/interactiveshell.py:3169: DtypeWarning:Columns (14,20,37,38) have mixed types.Specify dtype option on import or set low_memory=False.
/opt/homebrew/lib/python3.9/site-packages/IPython/core/formatters.py:918: UserWarning:
Unexpected error in rendering Lux widget and recommendations. Falling back to Pandas display.
Please report the following issue on Github: https://github.com/lux-org/lux/issues 

/opt/homebrew/lib/python3.9/site-packages/lux/core/frame.py:628: UserWarning:Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.9/site-packages/lux/core/frame.py", line 590, in _ipython_display_
    self.maintain_recs()
  File "/opt/homebrew/lib/python3.9/site-packages/lux/core/frame.py", line 432, in maintain_recs
    custom_action_collection = custom_actions(rec_df)
  File "/opt/homebrew/lib/python3.9/site-packages/lux/action/custom.py", line 76, in custom_actions
    recommendation = lux.config.actions[action_name].action(ldf)
[...]

PR #448, opened 2022-01-27
FIX-#403: Fixes mixed-dtype problem in execute_2D_binning

cgarciae:
## Overview

Mixed dtype columns cause `execute_2D_binning` to crash when trying to bin via `pd.cut`, this error shows up frequently in users code (#403, #394).

## Changes

* Adds a test using mixed dtypes
* Tries to cast series to `float` inside `execute_2D_binning`

### DIFF
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -395,16 +395,20 @@ class PandasExecutor(Executor):
             y_attr = vis.get_attr_by_channel("y")[0].attribute
 
             if vis.data[x_attr].dtype == np.dtype('O'):
-                try:
-                    vis.data[x_attr] = vis.data[x_attr].astype(float)
-                except ValueError:
-                    pass
+                mixed_dtype = len(set(type(val) for val in vis.data[x_attr])) >= 2
+                if mixed_dtype:
+                    try:
+                        vis.data[x_attr] = vis.data[x_attr].astype(float)
+                    except ValueError:
+                        pass
 
             if vis.data[y_attr].dtype == np.dtype('O'):
-                try:
-                    vis.data[y_attr] = vis.data[y_attr].astype(float)
-                except ValueError:
-                    pass
+                mixed_dtype = len(set(type(val) for val in vis.data[y_attr])) >= 2
+                if mixed_dtype:
+                    try:
+                        vis.data[y_attr] = vis.data[y_attr].astype(float)
+                    except ValueError:
+                        pass
 
             vis._vis_data["xBin"] = pd.cut(vis._vis_data[x_attr], bins=lux.config.heatmap_bin_size)
             vis._vis_data["yBin"] = pd.cut(vis._vis_data[y_attr], bins=lux.config.heatmap_bin_size)
--- a/lux/executor/PandasExecutor.py
+++ b/lux/executor/PandasExecutor.py
@@ -394,17 +394,17 @@ class PandasExecutor(Executor):
             x_attr = vis.get_attr_by_channel("x")[0].attribute
             y_attr = vis.get_attr_by_channel("y")[0].attribute
 
-            def _try_cast_float(series):
-                try:
-                    return series.astype(float)
-                except ValueError:
-                    return series
-
             if vis.data[x_attr].dtype == np.dtype('O'):
-                vis.data[x_attr] = _try_cast_float(vis.data[x_attr])
-            
+                try:
+                    vis.data[x_attr] = vis.data[x_attr].astype(float)
+                except ValueError:
+                    pass
+
             if vis.data[y_attr].dtype == np.dtype('O'):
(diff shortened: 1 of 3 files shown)


lux/executor/PandasExecutor.py
```diff
@@ -394,6 +394,18 @@ def execute_2D_binning(vis: Vis) -> None:
             x_attr = vis.get_attr_by_channel("x")[0].attribute
             y_attr = vis.get_attr_by_channel("y")[0].attribute
 
+            if vis.data[x_attr].dtype == np.dtype('O'):
+                try:
+                    vis.data[x_attr] = vis.data[x_attr].astype(float)
+                except ValueError:
+                    pass
```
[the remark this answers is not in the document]
dorisjlee: No we probably should not raise an exception here because general string columns are also recognized as object type, so we don't want to throw an error in those cases.

lux/executor/PandasExecutor.py
```diff
@@ -394,6 +394,18 @@ def execute_2D_binning(vis: Vis) -> None:
             x_attr = vis.get_attr_by_channel("x")[0].attribute
             y_attr = vis.get_attr_by_channel("y")[0].attribute
 
+            if vis.data[x_attr].dtype == np.dtype('O'):
```
dorisjlee, line 397: This is not a great detection mechanism for mixed type. There's a lot of cases where the column is properly detected as object (e.g. string). See [here](https://github.com/lux-org/lux/blob/master/lux/core/series.py#L151) for a better detection mechanism.

lux/executor/PandasExecutor.py
```diff
@@ -394,6 +394,18 @@ def execute_2D_binning(vis: Vis) -> None:
...
+                    vis.data[x_attr] = vis.data[x_attr].astype(float)
+                except ValueError:
+                    pass
+
+            if vis.data[y_attr].dtype == np.dtype('O'):
+                try:
+                    vis.data[y_attr] = vis.data[y_attr].astype(float)
+                except ValueError:
+                    pass
+
```
dorisjlee, line 408: How does this conversion interact with the warning we show [here](https://github.com/lux-org/lux/blob/master/lux/executor/PandasExecutor.py#L257)? For performance reasons, should we be doing this outside, as a more general dataframe sanitization step, instead of within specific execute visualization calls, so that the conversion is only done once per column, instead of every time per visualization?

tests/test_pandas.py
```diff
@@ -131,3 +131,13 @@ def test_datetime_index_serialize():
         df = pd.read_csv(tmpdir / "test.csv", names=["date", "c1"], index_col="date", header=0)
 
     df._ipython_display_()
+
+
+def test_cut_mixed_types():
+
+    a = [1, "2", 3, "4", 5, "6", 7, "8", 9, "10"] * 1_000
```
dorisjlee, line 138:
```suggestion
    a = [1, "2", 3, "4", 5, "6", 7, "8", 9, "10"] * 1000
```

tests/test_pandas.py
```diff
@@ -131,3 +131,13 @@ def test_datetime_index_serialize():
         df = pd.read_csv(tmpdir / "test.csv", names=["date", "c1"], index_col="date", header=0)
 
     df._ipython_display_()
+
+
+def test_cut_mixed_types():
```
dorisjlee, line 136: See also [this test](https://github.com/lux-org/lux/blob/master/tests/test_pandas_coverage.py#L669).

lux/executor/PandasExecutor.py
cgarciae: > How does this conversion interact with the warning we show [here](https://github.com/lux-org/lux/blob/master/lux/executor/PandasExecutor.py#L257)?

I don't see that warning about mixed types so it not reaching that path.

> For performance reasons, should we be doing this outside, as a more general dataframe sanitization step, instead of within specific execute visualization calls, so that the conversion is only done once per column, instead of every time per visualization?

I do agree, this type of visualization-specific sanitation happened in a previous issue as well and its not ideal. It seems we need a centralize sanitation mechanism that doesn't exist yet. I can open an issue for this.

dorisjlee: Sure let's open a separate issue for this.

lux/executor/PandasExecutor.py
dorisjlee: This is probably okay for now since this code only goes through 2D binning which is generally all quantitative values to begin with.

dorisjlee: Turns out the object type check is a more optimized approach than [pandas.api.types.is_object_dtype](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_object_dtype.html), so we're all good on this front!
Screen Shot 2022-02-07 at 7 13 29 AM

dorisjlee (approved):

Issue #447, opened 2022-01-21
Improve installation instructions / debugging UX


cgarciae:
* Improve readme
* Add warning if lux-widget extension is not installed.
  * Check if extension is enabled. 
* Add a `lux.debug_info` function that returns relevant package versions.

dorisjlee:
Resolved in #457.

PR #450, opened 2022-01-27
FIX-#447: Add show_versions helper for debugging

cgarciae:
## Overview

Adds a `show_versions()` function that prints the lux, pandas, and luxwidget versions. We can ask for this information for debugging.

PR #451, opened 2022-01-28
FIX-#447: Add debug information if luxwidget extension is not enabled.

cgarciae:
## Overview

Per #447, the idea is to add some debug information (i.e. a warning) in case `luxwidget` is not installed/enabled.

### DIFF
--- a/lux/utils/debug_utils.py
+++ b/lux/utils/debug_utils.py
@@ -1,3 +1,10 @@
+import json
+from pathlib import Path
+import re
+import subprocess
+import typing as tp
+
+
 def show_versions():
     """
     Prints the versions of the principal packages used by Lux for debugging purposes.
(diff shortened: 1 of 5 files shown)


lux/utils/debug_utils.py
```diff
@@ -22,5 +29,94 @@ def show_versions():
...
+
+    if status_str != "OK":
+        return False, f"luxwidget is installed but currently '{status_str}'"
+
+    return True, ""
+
+
+def is_lab_notebook():
+    import re
+    import psutil
```
dorisjlee, line 90: Note that we need to add psutil on requirements.txt. It seems like psutil is not something that comes as a default installation.

Screen Shot 2022-01-30 at 10 03 49 AM

lux/utils/debug_utils.py
```diff
@@ -0,0 +1,122 @@
...
+    ip = IPython.get_ipython()
+
+    # return if the shell is not available
+    if ip is None:
+        return
+
+    if is_lab_notebook():
+        enabled, msg = lab_enabled()
+    else:
+        enabled, msg = notebook_enabled()
```
dorisjlee, line 111: Note here, nothing is printed if the luxwidget is enabled. We should definitely send a confirmation printing out what lab/notebook version is used and check if the widget is enabled.

lux/utils/debug_utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # get the ipython shell
+    import IPython
+
+    ip = IPython.get_ipython()
+
+    # return if the shell is not available
+    if ip is None:
+        return
+
```
dorisjlee, line 107: Display the result of `is_lab_notebook` to users.

dorisjlee: Listing `jupyter --version` here would also be helpful.

lux/utils/debug_utils.py
cgarciae: Ah interesting, thanks! It was somehow on my env. I'll add it.

lux/utils/debug_utils.py
cgarciae: @dorisjlee the original idea was just to print the warning if a problem was detected.
You want to print out some info always?

dorisjlee: Yeah, if the user is explictly calling the debug function. It would be good to print out some verification/check information even if everything is working as expected. This would help with debugging what else might have gone wrong.

lux/utils/debug_utils.py
dorisjlee: Looks like jupyter version is captured in #450 already!

dorisjlee:
These commits were created in a separate branch and does not contain commit signatures. I'm closing this issue and migrating the code changes to the new PR #457 .

Issue #298, opened 2021-03-04
[BUG] Long filter values needs to be abbreviated


dorisjlee:
**Describe the bug**
Filters titles with very long values are not abbreviated, resulting in very wide visualizations.

**To Reproduce**
```python
from lux.vis.Vis import Vis
long_content = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."

dataset = [
    {"long_attr":long_content, "normal": 3, "normal2": 1},
    {"long_attr":long_content, "normal": 3, "normal2": 1},
    {"long_attr":long_content, "normal": 2, "normal2": 1},
    {"long_attr":long_content, "normal": 4, "normal2": 1}
]
test = pd.DataFrame(dataset)
vis = Vis(["normal2", "normal",f"long_attr={long_content}"], test)
```
![image](https://user-images.githubusercontent.com/5554675/109974153-396abd00-7d34-11eb-9ac6-a2121d84cf91.png)

This is also an issue for matplotlib but the results are cutoff: 
![image](https://user-images.githubusercontent.com/5554675/109974080-2657ed00-7d34-11eb-9091-9971ae21b711.png)

**Expected behavior**
We should abbreviate the values on filter. We are in fact doing abbreviation for attribute values and legend. Perhaps there is some shared code that we can reuse across these different abbreviation usages?

PR #465, opened 2022-03-08
FIX-#298: Abbrev long filter values and add test

labanyamukhopadhyay:
## Overview

If a title is longer than 25 characters (due to a long filter value), it is abbreviated so the visualization is not exceedingly wide or have a title that is cut off. Referencing #298  

## Changes

The long text is truncated in both Altair and Matplotlib charts so that chart_title = chart_title[:15] + "..." + chart_title[-10:].
tests/test_vis.py also includes test_abbrev_title to test for this behavior.

## Example Output
Screen Shot 2022-03-08 at 12 19 43 PM

### DIFF
--- a/lux/vislib/altair/AltairChart.py
+++ b/lux/vislib/altair/AltairChart.py
@@ -117,6 +117,8 @@ class AltairChart:
     def add_title(self):
         chart_title = self.vis.title
         if chart_title:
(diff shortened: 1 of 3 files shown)


labanyamukhopadhyay:
#298

PR #466, opened 2022-03-14
FIX-#298: Abbrev long filter values and add to vis test

labanyamukhopadhyay:
Signed-off-by: Labanya Mukhopadhyay <labanya@ponder.io>

## Overview

If a title is longer than 25 characters (due to a long filter value), it is abbreviated so the visualization is not exceedingly wide or have a title that is cut off. Referencing https://github.com/lux-org/lux/issues/298

## Changes

The long text is truncated in both Altair and Matplotlib charts so that chart_title = chart_title[:15] + "..." + chart_title[-10:].
tests/test_vis.py also includes test_abbrev_title to test for this behavior.

## Example Output
vis_test

### DIFF
--- a/tests/test_vis.py
+++ b/tests/test_vis.py
@@ -582,3 +582,23 @@ def test_intent_override_all_column():
     assert (
         "y = alt.Y('Record', type= 'quantitative', title='Number of Records'" in current_vis_code
     ), "All column not overriden by intent"
+
+
+def test_abbrev_title():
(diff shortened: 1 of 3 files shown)


tests/test_vis.py
```diff
@@ -582,3 +582,23 @@ def test_intent_override_all_column():
...
+    df = pd.DataFrame(dataset)
+    lux.config.plotting_backend = "matplotlib"
+    vis = Vis(["normal2", "normal", f"long_attr={long_content}"], df)
+    vis_code = vis.to_matplotlib()
+    print(vis_code)
+    assert "long_attr = Lor...t laborum.'" in vis_code
+
+    vis_code = vis.to_altair()
+    print(vis_code)
+    assert "long_attr = Lor...t laborum.'" in vis_code
```
dorisjlee, line 604:
```suggestion
    assert "long_attr = Lor...t laborum.'" in vis_code
    lux.config.plotting_backend = "altair"
```

dorisjlee: The sql test errors seem to be resulting from the fact that matplotlib plotting backend is set here and used throughout. Note that the "lux" variable persists throughout the entire test session, so any configs should set this back to the default value, so that it doesn't affect other tests suites outside of this function.

dorisjlee:
Nice work, thanks @labanyamukhopadhyay!

Issue #493, opened 2023-06-19
[Feature Request] Lux should also be extensible to make it easy for external libraries with pandas API to support.


rentruewang:
**Is your feature request related to a problem? Please describe.**
[Modin](https://github.com/modin-project/modin) is a drop in replacement for pandas. However, current version of lux doesn't work with it.

**Describe the solution you'd like**
Lux should work with modin and any library with pandas API.

**Describe alternatives you've considered**
Refactor lux's logic into mixins / use factory patterns to construct dataframes. This way, it'll be easy to extend lux.

PR #494, opened 2023-06-19
FEAT-#493: Extensible.

rentruewang:
After this refactor, Making `Vis` etc support the pandas library is as easy as defining 4 1-liner classes.

## Overview

For example, for modin defining this one-liner class is enough for it to be used with `Vis`:

```python
class LuxModinDF(LuxDataFrameMixin, modin.DataFrame): pass
```

then wrap the dataframe `df` by:

```python
df = LuxModinDF(df)
```

## Changes

I refactored Lux's pandas subclasses into mixins and thin wrappers using mixins (much like the above one liner).

## Example Output

No change.

### DIFF
--- a/lux/core/__init__.py
+++ b/lux/core/__init__.py
@@ -70,4 +70,4 @@ def setOption(overridePandas=True):
         pd.Series = originalSeries
 
 
-setOption(overridePandas=True)
+setOption(overridePandas=False)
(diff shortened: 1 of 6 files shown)


dorisjlee:
@rentruewang can you run the linter to prevent test fail?

rentruewang:
@dorisjlee Do you know what linter I should run? I was able to pass pre-commit, and I didn't make changes to those that failed `black --check`. Would you like me to run `black` on those files?

